PR #17, opened 2019-07-18
Added annotation properties for ELAN generation.

mpomarlan:
Adds two fairly generic annotation properties, and for each of them a more specific subproperty related to a particular user group of the ontology.

"nickname": similar to 'label', but without GUI or reasoner impact in Protege; avoiding Protege problems is why this is not a label subproperty. Nicknames are used by particular software working with the ontology as a way to give community-specific names.

"ELANName": subproperty of nickname. If present, used by the controlled vocabulary generation script to create a vocabulary entry. By default, the concept name itself is used for the vocabulary entry.

"UsageGuideline": annotation of a concept with an indication of how it is to be used in some context.

"ELANUsageGuideline": provides a short description, similar to a definition, of a controlled vocabulary entry for ELAN. Is a subproperty of UsageGuideline.

### DIFF
--- a/owl/EASE.owl
+++ b/owl/EASE.owl
@@ -24,6 +24,34 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE.owl#ELANName -->
+
+    <owl:AnnotationProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#ELANName">
+        <rdfs:subPropertyOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#nickname"/>
+    </owl:AnnotationProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE.owl#ELANUsageGuideline -->
+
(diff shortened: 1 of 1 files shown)


daniel86:
what about adding rdfs:comment for the properties? They are kind of self-explanatory. Should we insist on every concept and property having a rdfs:comment?

mpomarlan:
Sure, I can add rdfs:comment to these properties.

About their systematicity, a step in that direction, I hope, are the more general versions of the properties; that is, I don't just have ELANx property, but also something larger, more general, that intends to capture something useful about annotations targeted at a specific purpose.

Another little issue is superproperties of nickname (I'm fairly certain this should not be label, for Protege confusion reasons) and UsageGuideline. For UsageGuideline, the default comment can be a superproperty and at least the ECV generation would not be confused. What do you think?

daniel86:
@mpomarlan I agree that we should not use *rdfs:label*, and probably also avoid using the phrase label due to Protege using it in a special way. Else I do not see a problem with annotation property hierarchies in case Protege allows it.

PR #20, opened 2019-07-18
adjusted description of *DesignedBlade*

### LINKED ISSUE #18: Blade description
Currently, the description of DesignedBlade is

<rdfs:comment>An item designed to cut things.</rdfs:comment>

It used to be (and in the ELAN branch still is)

<rdfs:comment>An item designed to provide at least one sharp edge.</rdfs:comment>

which I think is preferable. The characteristic of the blade is that it has at least one sharp edge. It is true that basically all blades are used for cutting-- I say basically all out of charity, I can't think of a blade not meant to cut-- however, imo, it is not true that all items designed for cutting are, or have, blades.  A counterexample is a laser or waterjet cutter, and a cheese-cutting string would not be considered a blade either.

daniel86:
fixes #18


daniel86:
@mpomarlan fixed typo, could you approve now?

PR #25, opened 2019-08-04
+6DPose +3DPosition +EuclideanPoint

daniel86:
First attempt to model regions in euclidean space.
In ROS, 6D poses (translation + quaternion) are the most common way to locate things in space.

Note, that xsd:string is used for encoding vector data type properties, since xsd does not define vector types and I want to avoid having too many data triples that are not used anywhere.
This is a bit ugly. Two other options would be defining vector types, or second doing one assertion for each component. vector types would reduce number of assertions while making it impossible with standard tools to access the component values (KnowRob would understand them). Do you think we should use the many assertions variant instead?


porzel:
Having looked at the class "Abstract" recently, is seems this would be the place to model such data types.

owl/EASE.owl
```diff
@@ -1144,6 +1189,42 @@ The second issue relates to what is &quot;outside&quot;. The state is a configur
...
 
 
+    <!-- http://www.ease-crc.org/ont/EASE.owl#3DPosition -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE.owl#3DPosition">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#EuclideanPoint"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE.owl#hasTranslation"/>
+                <owl:someValuesFrom rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
mpomarlan: Perhaps this should be an exactly 1 restriction, rather than some, because a 3D position will have exactly one translation. There is of course a question of which reference frame, and which unit of measurement/scale, this translation is expressed in. However, hasTranslation is a data property to a string, and it's not explained that this string should also contain reference frame and scale.

Alternatively, several translations for a point may indicate some probabilistic approach, where a point is described by a collection of "particles", where each particle mass corresponds to probability. Again though, this isn't explained.

I'm giving the examples as a suggestion that there are cases where it is useful to store more translations for the same point, but we should explain if we have such a situation in mind, and if not, I think we can use the exactly 1 restriction.

owl/EASE.owl
```diff
@@ -1144,6 +1189,42 @@ The second issue relates to what is &quot;outside&quot;. The state is a configur
...
+        </rdfs:subClassOf>
+        <rdfs:comment>A point in three dimensional space, given as translation.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE.owl#6DPose -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE.owl#6DPose">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#EuclideanPoint"/>
```
mpomarlan: Euclidean points, and points in general, are often not regarded as having an orientation. This is because usually when one speaks of a point in Euclidean space, they mean a linear space such as R^3, not the special Euclidean group SE(3).

Typically, a point might be regarded as a part of some orientable mathematical object.

owl/EASE.owl
```diff
@@ -1144,6 +1189,42 @@ The second issue relates to what is &quot;outside&quot;. The state is a configur
...
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE.owl#hasRotation"/>
+                <owl:someValuesFrom rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
mpomarlan: As before, we should consider whether we need more rotations/translations here (and a "some" restriction) or whether we have an "exactly 1" restriction.

owl/EASE.owl
```diff
@@ -1144,6 +1189,42 @@ The second issue relates to what is &quot;outside&quot;. The state is a configur
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE.owl#hasTranslation"/>
+                <owl:someValuesFrom rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
mpomarlan: As before, it's not clear how we're using several values here, if we had them. Suggest an "exactly 1" restriction instead.

owl/EASE.owl
daniel86: Yes, the idea was that you could have assertions with different units of measurement. I will change it to exactly 1 for

owl/EASE.owl
daniel86: I was thinking more general, "a point is an element of some space".

daniel86: Well since you did not make a suggestion I will just remove the concept EuclideanPoint

owl/EASE.owl
```diff
@@ -852,6 +852,41 @@ Sub-properties are used to distinct between different cases of event endpoint re
...
+        <rdfs:label>has reference frame</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE.owl#hasRotation -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#hasRotation">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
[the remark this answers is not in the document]
daniel86: changed it to 'Space region' or 'Physical object', but not really happy with it

daniel86:
damn seems something from the other pull request slipped in :/ I will try to resolve that now

daniel86:
> damn seems something from the other pull request slipped in :/ I will try to resolve that now

fixed now. @mpomarlan  please have a look again

daniel86:
>  recently, is seems this would



> Having looked at the class "Abstract" recently, is seems this would be the place to model such data types.

Yes, it is already in the initial pull request since *Region* is a sub-class of *Abstract*. Or you rather mean *FormalEntity* ?

daniel86:
> > damn seems something from the other pull request slipped in :/ I will try to resolve that now
> 
> fixed now. @mpomarlan please have a look again

@mpomarlan ping

PR #26, opened 2019-08-04
+Shape +Mass +Friction

daniel86:
definitions to describe some regions of physical objects that I need for interfacing URDF.

Note that I considered to use qudt to explicate units (e.g. kilogram), but it seems that the current version does not work with OWL, but it seems version 2 will soon be released.


owl/EASE-OBJ.owl
```diff
@@ -26,6 +26,17 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasShape -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasShape">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRegion"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity may be too wide a domain for hasShape.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+     -->
+
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasDepth -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasDepth">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasDepth.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has depth</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFilePath -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFilePath">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasFilePath

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has file path</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasHeight -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasHeight">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasHeight.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has height</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasLength -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasLength">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasLength.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has length</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasRadius -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasRadius">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasRadius.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has radius</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasWidth -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasWidth">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasWidth.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has width</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasXScale -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasXScale">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasXScale.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has x scale</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasYScale -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasYScale">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasYScale.

owl/EASE-OBJ.owl
```diff
@@ -79,6 +90,116 @@
...
+        <rdfs:label>has y scale</rdfs:label>
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasZScale -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasZScale">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
```
mpomarlan: Entity is too large a domain for hasZScale.

owl/EASE-OBJ.owl
```diff
@@ -178,6 +326,27 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.w3.org/2001/XMLSchema#float"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasRadius"/>
+                <owl:someValuesFrom rdf:resource="http://www.w3.org/2001/XMLSchema#float"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A solid geometrical figure with straight parallel sides and a circular or oval cross section.</rdfs:comment>
```
mpomarlan: If oval, it should have two radii corresponding to the oval's major and minor axis.

owl/EASE-OBJ.owl
```diff
@@ -449,6 +651,15 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Shape -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Shape">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Region"/>
```
mpomarlan: Given that Cylinder etc. are subclasses of Shape, it is certainly appropriate to have Shape be a subclass of Region. What is the Quality that Shape is the Region for?

daniel86: Appearance for example. But in the case of URDF the visual or the collision model

owl/EASE-OBJ.owl
daniel86: some values allows having multiple radii, so no problem

owl/EASE-OBJ.owl
daniel86: I am not sure about being super restrictive here.
What is your suggestion?

First, qualities may refer to regions.
Is this exclusively? i.e. a region may only be the region of some quality concept but not the other? I do not think so.
Take the example of *Kinematics* quality which is kind of a collection of all attributes contributing to kinematics. But these attributes may contribute to other qualities as well.
So I would not be more specific then saying any quality may take this region.

Second, objects may also refer to regions.
i.e. DUL allows to directly relate an Object to some Region. You could say that only PhysicalObjects have a shape, maybe. But what about e.g depictions of social objects?

Last, events may also have regions. Can any event have a shape? Well it could! Like when people are moving around, the bounding box of their movement for example.

owl/EASE-OBJ.owl
daniel86: same as above, and same for the identical comments below.

owl/EASE-OBJ.owl
mpomarlan: Likewise any event can have a color. With metaphor you can do anything.

But we're not dealing in metaphors. A shape is a particular kind of thing, and it does make sense to limit it to Physical Objects. A social object cannot have a shape. A physical depiction of it does. I don't see where having shapes for events will prove useful.

owl/EASE-OBJ.owl
mpomarlan: I'm not moved by that answer. Width, depth, length are here physical dimensions of physical objects. This is not about construing things metaphorically.

owl/EASE-OBJ.owl
daniel86: No it is not a metaphor. The event can have a quality, like the space region in which it took place, which is forming a shape. No metaphor at all

daniel86: "Likewise any event can have a color" -- what about saying that events may have qualities that have color regions? that seems to be fine to me

owl/EASE-OBJ.owl
mpomarlan: Then consider "max 2" when dealing with ellipses, and "max 3" when dealing with ellipsoids.

daniel86: I would suggest to have subclasses that specify this

owl/EASE-OBJ.owl
mpomarlan: Qualities definitely must be in the domain of any property with range Region, so there's no objection there.

daniel86: But qualities are generally optional in DUL. So you can relate Event/Object directly to regions without going via Quality. Hence, I used Entity as domain

mpomarlan: The more things we allow the less the ontology can tell us; for example, if X hasShape something, you know nothing of X if the domain of hasShape is Entity.

Perhaps this is sometimes appropriate, but for an ontology of everyday activity it seems plausible that we can consider only physical objects, and their appearances, to have shapes.

daniel86: But what does "X is a quality" buy us? Or it could be "X is a (physical) object or quality", also does not seem to be helpful.

More useful are axioms at the object concepts, that restrict their qualities, I think. But the property domain of hasRegion sub-properties does not seem to be super relevant for reasoning, or am I missing something?

mpomarlan: I would say this is a general principle, to try and think of as many restrictions as will not hurt the class of models, and put them in. It is easy to not specify a domain for a property, but I think in our case we can specify a domain without missing out on any useful models for our application, and therefore we should.

daniel86: So what is your suggestion?

mpomarlan: PhysicalObject and Appearance (the corresponding quality of shape).

daniel86: I am trying to follow principles of DUL here. And as explained before, DUL allows to put regions on objects and events. I *was* thinking about this before, and decided that this would be in the sense of DUL

daniel86: > PhysicalObject and Appearance (the corresponding quality of shape).

I am still questioning if there is one **corresponding** quality. I was arguing before that attributes may contribute to multiple qualities. And the shape may not only contribute to appearance

mpomarlan: "Contribute" is also vague. Mass is a quality; it should also correspond to a region in positive real number space, or some qualitative space on top of such. Many things can contribute to Mass, but it would not be appropriate to say that Mass takes values in a region for material density. That's not what Mass is.

daniel86: Mass is not a quality in DUL, it is a *PhysicalAttribute*

daniel86: I assume you know what I meant with contribute. that should be obvious: i.e. that it's a region of that quality.

mpomarlan: No actually I don't, since we're debating what I thought obvious, that only Physical Objects and the appearance quality should have a shape.

daniel86: ok sorry for being unclear

daniel86: So regarding that, we would need to clarify if there is a 1-to-1 or 1-to-many mapping between Regions and Qualities. So the question can a region be the region of two distinct qualities?
It is hard to answer this question without having a complete taxonomy of qualities in DUL. Actually that makes it hard to think about what are all the qualities that are out there

daniel86: So thinking of Qualities as *qualifications* of quantitative regions, I think it makes sense that the same region can be qualified in different ways, generally

PR #28, opened 2019-08-11
+Shape +Mass +Friction

daniel86:
second try...
Shape, Mass, Friction defined as physical attributes which are regions. Note that DUL enforces that qualities are not used at all (i.e. *PhysicalAttribute is-region-of only PhysicalObject*).


porzel:
Seems OK.

porzel:
Please use the naming converntion *Region or *Attribute for subregions, e.g. densitiyRegion or densityAttribute and leave the term Density for the Quality (if we ever need/want to model them as well.

daniel86:
> Please use the naming converntion *Region or *Attribute for subregions, e.g. densitiyRegion or densityAttribute and leave the term Density for the Quality (if we ever need/want to model them as well.

done

daniel86:
> > Please use the naming converntion *Region or *Attribute for subregions, e.g. densitiyRegion or densityAttribute and leave the term Density for the Quality (if we ever need/want to model them as well.
> 
> done

oh I forgot the properties

daniel86:
> > > Please use the naming converntion *Region or *Attribute for subregions, e.g. densitiyRegion or densityAttribute and leave the term Density for the Quality (if we ever need/want to model them as well.
> > 
> > 
> > done
> 
> oh I forgot the properties

now its done

PR #30, opened 2019-08-27
Execution state

daniel86:
a region with labels identifying the state of execution


daniel86:
To be honest, the listing came from the top of my head, I haven't checked literature. But Sebastian will try to map CRAM action errors to this enum

PR #31, opened 2019-10-24
Object Qualities + Software Diagnosis

daniel86:
These are changes I made to make parts of KnowRob work with ease_ontology.

First, I need to have some object qualities, and a way to handle quantitative data.
e.g. "an object has a color which is encoded in RGBA floating point values, and these values are [R,G,B,A]".
Second, I added some definitions to EASE-IO for software diagnosis.

Sorry that I did not separate them, I hope a review is still possible.
Else let me know and I'll do multiple requests.


owl/EASE-IO.owl
```diff
@@ -33,16 +33,27 @@
...
             <owl:Class>
                 <owl:intersectionOf rdf:parseType="Collection">
-                    <rdf:Description rdf:about="http://www.ease-crc.org/ont/EASE-IO.owl#DigitalObject"/>
-                    <rdf:Description rdf:about="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Agent"/>
+                    <rdf:Description rdf:about="http://www.ease-crc.org/ont/EASE-IO.owl#SoftwareImplementation"/>
+                    <rdf:Description rdf:about="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalAgent"/>
                 </owl:intersectionOf>
             </owl:Class>
         </owl:equivalentClass>
+        <rdfs:comment>An agent that is implemented in software.</rdfs:comment>
```
mpomarlan: I think I like the original approach of having ComputationalAgent be an Agent, but not necessarily a Physical one.

Is Alexa a ComputationalAgent? Intuitively yes. Is it a Physical one? This is tricky, as I can see people classing it either way.

In any case, if a computational agent must be physical, then it cannot only be implemented in software. There needs to be a hardware implementation to support it.

owl/EASE-IO.owl
```diff
@@ -51,6 +62,69 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-IO.owl#DigitalObject"/>
+        <rdfs:comment>A structured archive of software that also often contains data needed to run the software, and scripts to build and configure the software.</rdfs:comment>
+        <rdfs:label>Software package</rdfs:label>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-IO.owl#SoftwareType -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-IO.owl#SoftwareType">
```
mpomarlan: This is very nice!

owl/EASE.owl
```diff
@@ -1126,14 +1055,6 @@ Such an event does NOT fall under the definition of Accident here. An example of
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE.owl#Location -->
```
mpomarlan: What's the current plan about ontologically describing locations?

owl/EASE-IO.owl
daniel86: My argument was that the agent allocates memory on the computer device, and that the space region in the memory where the agent is represented is a physical manifestation of the agent. I do not insist on that, but I definitely need *InformationRealization* concepts with object character such that they can participate in events.
That's why I introduced *DigitalObject* which is a super-class of *ComputationalAgent*.
Digital object already is subclass of *PhysicalObject*, so in case we do not want to commit to computational agents being physical objects, we also need to touch this definition.

My motivation comes from IOLite where a distinction is made between *InformationObject* and *InformationRealization*.
IOLite comments that one example is the conceptualization of laws in some context (InformationObject) vs. a book with the laws written down (InformationRealization).
In the digital world we have *DataStructure* (InformationObject) vs. *DigitalResource* (InformationRealization).

Here, I would see *ComputationalAgent* analogous to *Robot* -- an entity that can solve some problems when appropriate input is specified, and when it is properly installed and running. In other words some piece of code that is currently executed on a computer, and that has an associated space region (however unknown) in the memory of the computer.
However, more distinctions could be possible like an installed but not running agent that also takes some space on the hard drive

owl/EASE.owl
daniel86: Please have a look in EASE-OBJ, the concept should have moved there.

Qualitatively, objects may have a *Localization* which is a quality, it takes values only from *SpaceRegion*. Currently, we have 6DPose and 3DPosition underneath *SpaceRegion*. This is not yet very fine-grained. I struggled a bit with getting a more fine grained taxonomy below it, but maybe you can give it a try at some point! For now, I was thinking it is sufficient to have this shallow taxonomy.

As *SpaceRegion*, KnowRob will rather store the data into mongo DB instead of having it in the triple store. This allows for changing the regions of a quality over time without having to deal with problems regarding representing time in OWL.

daniel86: Oh and regarding quaternions: I doubt that we will ever qualify quaternion components, so I removed the class in favour of the dataproperty *has orientation vector*

owl/EASE-IO.owl
daniel86: Still hard to come up with a good taxonomy! I tried already, the current result is in the *SRDL* package of KnowRob. But I think this will need a couple of iterations to only cover software of robot control systems.

owl/EASE-IO.owl
mpomarlan: Ok.

Then a minor point on the rdfs:comment content: should it not be important that a computational agent be implemented in hardware as well? Or, if implemented is not the right relation, then bring hardware into the description somehow?

I ask because it seems to me that a casual use of implemented wouldn't necessarily imply your concept about how a computational agent takes up physical resources to exist and then be properly installed and running to act.

daniel86: I remove the physical stuff now, we can discuss this issue at a later point

daniel86:
Please have a look again @mpomarlan

PR #34, opened 2019-10-28
An initial crack at modelling dispositions and affordances.

mpomarlan:
This pull request is to encourage/provide a forum for a written discussion on the disposition and affordance modelling.

I have written what we have so far about the abstract level of Dispositions, made some changes to the concrete examples involving containment, and defined a situation and affordance for containment.

There are several things left to do, such as:
-- what was the connection between Agent and Affordance supposed to be again?
-- modelling trigger and background for either dispositions or affordance manifestations.

### DIFF
--- a/owl/EASE-OBJ.owl
+++ b/owl/EASE-OBJ.owl
@@ -50,6 +50,24 @@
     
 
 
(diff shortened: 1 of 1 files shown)


mpomarlan:
I'll close this since we have the disposition model in the main branch.

PR #36, opened 2019-11-07
iai_kitchen

daniel86:
- populated EASE-OBJ with furniture in IAI kitchen
- initial taxonomy of roles
- initial taxonomy of qualities
- disposition taxonomy as suggested by Robert with slight changes
- finally deleted EASE-middle.owl


owl/EASE-OBJ.owl
```diff
@@ -26,6 +26,46 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#affects -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#affects">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasConstituent"/>
```
daniel86: note: I have not seen a property that would be a better match. e.g. 'is related to concept' is only concept-to-concept.

owl/EASE-OBJ.owl
```diff
@@ -26,6 +26,46 @@
...
+        <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Disposition"/>
+        <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:comment>Associates a disposition to a role affected by it.</rdfs:comment>
+    </owl:ObjectProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#attracts -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#attracts">
```
daniel86: the attracts chain is useful to axiomatize objects that have some disposition.

owl/EASE-OBJ.owl
```diff
@@ -204,17 +372,6 @@
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE.owl#hasDesignatedPatient -->
-
-    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#hasDesignatedPatient">
```
daniel86: this is not needed any more, now using the *affects* property

owl/EASE-OBJ.owl
```diff
@@ -526,10 +779,55 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#CleanlinessRegion"/>
+        <rdfs:comment>A cleanliness region with values considered as clean.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Cleanliness -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Cleanliness">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#SocialQuality"/>
```
daniel86: this is maybe a bit questionable, my rationale was that this quality is defined within a social context.

daniel86:
@mpomarlan I couldn't merge with the ease-disposition branch. Please have a look whether you would like to merge manually.

owl/EASE-OBJ.owl
```diff
@@ -144,6 +276,19 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#isModifiedBy -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#isModifiedBy">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isConstituentOf"/>
```
mpomarlan: Logically, this does not result in contradictions or concepts that are subconcepts of Nothing, so on that front it's ok.

On the other hand, this appears to insert a relational aspect "too soon" in the disposition model. Example:

I have a spraycan of paint. I can use it to alter the color of a wall, so presumably it will have some sort of Alteration disposition.

I think your intuition is, since I just described the spraycan by "I can alter the color of a wall", clearly I had to make reference to a modified quality in order to make sense of what the spraycan does.

My problem is, when I talk about the spraycan's disposition in general, I don't have a particular wall, with its particular Color quality individual, in mind. Yes, when we describe an event of actual painting, it will be useful to have a modifies/isModifiedBy relation, because the Event is about specific entities interacting.

However when discussing the disposition in the abstract, I don't want to depend on something conjectural and outside of it. That is, the constituency of the disposition of a spraycan cannot be the Color quality of a particular wall.

owl/EASE-OBJ.owl
```diff
@@ -438,11 +595,65 @@
...
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Alteration -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Alteration">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Disposition"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#modifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Quality"/>
```
mpomarlan: Is it some or is it only?

I can have a spraycan of paint. Clearly it has the disposition to paint a surface. However, I could store this spraycan, unused, until the end of time, and not have it participate in any painting event.

Qualities, also, should not be treated as abstract disembodied individuals. DUL's axioms are a bit weaker here in that they don't require a Quality to be "is quality of" something, but the DUL comment about qualities is clear that a Quality is "Any aspect of an Entity (but not a part of it), which cannot exist without that Entity."

So a spraycan of paint can, indeed, modify colors, but when we have in mind a Color individual we should also have in mind the Object (or whatever) individual that Color belongs to. So there cannot be an abstract statement that this spraycan has a disposition to modify some platonic Color individual. It can, in a particular event, be used to modify the color of a particular wall.

owl/EASE-OBJ.owl
```diff
@@ -438,11 +595,65 @@
...
+    </owl:Class>
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#AvoidedObject -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#AvoidedObject">
-        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
-        <rdfs:comment>An object that is avoided by the agent.</rdfs:comment>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#ExcludedObject"/>
```
mpomarlan: I wonder which of these sounds more natural:

I avoid X, therefore I exclude X
I exclude X, therefore I avoid X

I'm leaning towards the second-- exclusion makes clear that there is a particular region where X won't be, whereas avoidance just says I and X won't be in the same more or less exact place. In any case though, it's not too strong a lean either way, which suggests Avoidance and Exclusion are different things, not necessarily linked by subclassing.

owl/EASE-OBJ.owl
```diff
@@ -455,6 +666,54 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Barrier -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Barrier">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Excluder"/>
+        <rdfs:comment>A tool used as a barrier to prevent access to a protected space, or to prevent enclosed objects from escaping.</rdfs:comment>
```
mpomarlan: I see you either thought along similar lines to me or adapted some of the disposition branch I made into these commits :)

The reason I invented the Barrier concept though, as opposed to Robert's Inclusion/Exclusion, was because I didn't see that Containment and other similar schemas are about Exclusion or Inclusion.

If I want to prevent access to a protected space, I'm definitely excluding the stuff that's out. But if I want to prevent enclosed objects escaping, I'm not being exclusionary.

owl/EASE-OBJ.owl
```diff
@@ -455,6 +666,54 @@
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Blockage -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Blockage">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Exclusion"/>
```
mpomarlan: As before, while blockages are often used for exclusion, that's not always the case.

owl/EASE-OBJ.owl
mpomarlan: Actually that is probably exactly right.

One can objectively measure the number of contaminants etc, but one has to make a commitment that this is a contaminant whereas this other thing is ok to have. Such a decision is definitely the creation of an agent.

owl/EASE-OBJ.owl
```diff
@@ -571,50 +869,123 @@
...
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Containment -->
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Coverage -->
 
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Containment">
-        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Design"/>
-        <rdfs:comment>The description of a design that allows to contain other objects.</rdfs:comment>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Coverage">
```
mpomarlan: Coverage, like Containment, are more like barriers than strict exluders or includers. They can be used for both.

I can cover a pot to prevent flies from getting in. I can cover a pot to prevent popcorn from flying out. It's not very natural to refer to the second situation as being exclusionary.

owl/EASE-OBJ.owl
```diff
@@ -1004,28 +1796,58 @@
...
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#RemovedObject -->
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Regulation -->
 
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#RemovedObject">
-        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
-        <rdfs:comment>An object that is removed from something.</rdfs:comment>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Regulation">
```
mpomarlan: Yea, this works. Regulate, therefore Alter is a very intuitive chain.

owl/EASE-OBJ.owl
```diff
@@ -1236,34 +2319,49 @@ It is also possible that a Transient transitionsBack to the initial object. An e
...
-                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasDesign"/>
-                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Containment"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasQuality"/>
-                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Capacity"/>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasDisposition"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Enclosing"/>
```
mpomarlan: I didn't comment on the Includer/Excluder dispositions since those are fine on their own.

It's when they interact with schemas such as barrier, coverage, or containment that I am wary about committing to only one of them.

Containment can be protective (which is to say, exclusionary), or it can be to fix in place the contained (which is to say, inclusionary).

owl/EASE-OBJ.owl
```diff
@@ -1294,35 +2392,329 @@ It is also possible that a Transient transitionsBack to the initial object. An e
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Appliance"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#DesignedContainer"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasDisposition"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onClass>
+                    <owl:Class>
+                        <owl:intersectionOf rdf:parseType="Collection">
+                            <rdf:Description rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Enclosing"/>
```
mpomarlan: A dishwasher is an interesting case where containment of the tableware isn't really exclusionary nor inclusionary. The tableware isn't put inside the dishwasher because it would fly away otherwise (so, not inclusionary) nor is it put in there to protect them from outside stuff, necessarily. They simply are put in the same container where the water jets, who are contained inclusionarily (we don't want them to get out), operate.

owl/EASE-OBJ.owl
```diff
@@ -1294,35 +2392,329 @@ It is also possible that a Transient transitionsBack to the initial object. An e
...
+        </owl:equivalentClass>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#PhysicalEffector"/>
+        <rdfs:comment>An arm or leg of an embodied agent.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE.owl#PrehensileEffector -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE.owl#PrehensileEffector">
```
mpomarlan: huh. This looks oddly familiar. Was it in some other onto file or branch? I remember writing something like this concept once.

mpomarlan:
The TL;DR of the above is

-- the vast majority of this is ok, we just need to resolve a couple of issues that are foundational.
-- hasConstituent is a property that comes with rather hard commitments from the DUL axiomatization of Qualities (and I would agree with those commitments). It is not possible to have as constituents of a quality something which is not a quality
-- barriers are not necessarily excluders nor includers, and same for containers
-- a disposition should not have as constituent a quality it affects, because qualities are not platonic abstracts but belong to specific entities; on the other hand, the disposition does not need to manifest itself by affecting a particular object

owl/EASE-OBJ.owl
daniel86: but what should be the super-property? I haven't seen a good fit

owl/EASE-OBJ.owl
daniel86: yes it was somewhere else. maybe EASE-middle

owl/EASE-OBJ.owl
daniel86: I don't get what you write here. the tableware is enclosed, so it is included, right?

owl/EASE-OBJ.owl
daniel86: ok I see your point.

owl/EASE-OBJ.owl
daniel86: *attracts*: no, at least not intentionally. I think it is often ok to only have general properties and different contexts where the property is used. But well then the range/domain axioms must also be accordingly general which is not the case here. Hard to say if it should be more specific then Entity-Entity in case of a general attracts property.

owl/EASE-OBJ.owl
```diff
@@ -26,6 +26,46 @@
...
+        </owl:propertyChainAxiom>
+        <rdfs:comment>Associates a disposition to an object attracted by it.</rdfs:comment>
+    </owl:ObjectProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#disposes -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#disposes">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasConstituent"/>
```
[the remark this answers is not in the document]
daniel86: I will change the comments. About examples, yes I know. But there were so many new concepts. Let's add example later on

owl/EASE-OBJ.owl
daniel86: No, my reasoning is rather: I want to express what aspect of the object is altered, so I need to refer to the quality type. That's what we were talking about earlier... where I said that we may need to use *QualityType* here to classify qualities. But that seems silly at the moment. It was hard enough to come up with one quality taxonomy :)

It's strange that DUL does not assert that a quality must be linked to a object holding it! So it would not make the ontology inconsistent if dispositions have their individual qualities, as far as I can tell.
I was also thinking about only using universal restrictions which would also not cause problems in the ABOX, and could be used for reasoning in the TBOX.

So, any suggestion how to do it?

owl/EASE-OBJ.owl
daniel86: let's discuss this on Monday with @porzel

owl/EASE-OBJ.owl
daniel86: So to say spatially included within the frame of the dishwasher

owl/EASE-OBJ.owl
daniel86: I guess the taxonomy needs to move to the concept branch

owl/EASE-OBJ.owl
daniel86: only, you are right

owl/EASE-OBJ.owl
daniel86: yes, I was pointing you to this one during our last meeting. I will move it one level up then

owl/EASE-OBJ.owl
daniel86: I thought along similar lines :)

owl/EASE-OBJ.owl
daniel86: Quality taxonomy has moved to concept branch.

daniel86:
@mpomarlan should be fixed now, except of
-- barriers are not necessarily excluders nor includers, and same for containers

owl/EASE-OBJ.owl
daniel86: Well you could argue that you exclude the enclosed objects from the rest of the world, right? I was rather seeing *Inclusion* as "make an object part of a bigger whole" -- e.g. storing it together with other objects in a common compartment. And *Exclusion* as either prevent from entering, leaving or seeing some space.

owl/EASE-OBJ.owl
daniel86: The blocker prevents to move on certain pathways, but it only excludes objects that can not overcome the blockage -- e.g. you won't prevent dust particles from moving around there.

owl/EASE-OBJ.owl
daniel86: however, in the second situation we can say that the popcorn is enclosed by the pot and thus included, the cover of the pot clearly has an excluding component preventing the cold air to get inside to avoid the popcorn cooling down quickly. So I would argue the popcorn is included in the pot, and air is excluded to go inside through the cover.

owl/EASE-OBJ.owl
```diff
@@ -26,13 +26,209 @@
...
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasColor -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasColor">
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasQuality"/>
         <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isColorOf"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
         <rdfs:range rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Color"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Care with this property chain since I think we're running into similar problems of inconsistent properties.

So the range of hasColor is Color, which is perfectly fine. Color is an Extrinsic, which is a PhysicalQuality, which is a Quality.

According to the DUL axiomatization, the range of isClassifiedBy is Concept, which is a SocialObject, which is an Object.

Qualities and Objects are disjoint.

owl/EASE-OBJ.owl
```diff
@@ -69,6 +280,10 @@
         <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isLocalizationOf"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
         <rdfs:range rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Localization"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Just as in the case of hasColor, this appears similarly inconsistent with the DUL foundation. If a Localization is a Quality, then it cannot be in the range of isClassifiedBy, which is Concept.

owl/EASE-OBJ.owl
```diff
@@ -87,13 +302,30 @@
...
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasShape -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasShape">
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasQuality"/>
         <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isShapeOf"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
         <rdfs:range rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Shape"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Similar to the case of hasColor and hasLocalization. A Shape, which is an Intrinsic therefore a Quality, cannot be in the range of isClassifiedBy.

owl/EASE-OBJ.owl
```diff
@@ -120,36 +352,204 @@
...
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#isColorOf -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#isColorOf">
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isQualityOf"/>
         <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Color"/>
         <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Color, which is a Quality, cannot be in the domain of classifies, which is Concept.

owl/EASE-OBJ.owl
```diff
@@ -120,36 +352,204 @@
...
+    </owl:ObjectProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#isDesignOf -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#isDesignOf">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#describes"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Design"/>
+        <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Object"/>
```
mpomarlan: We should consider whether we can commit further, and require the range to be DesignedArtifact, but for the moment I'm ok with this.

owl/EASE-OBJ.owl
```diff
@@ -120,36 +352,204 @@
...
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#isLocalizationOf -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#isLocalizationOf">
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isQualityOf"/>
         <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Localization"/>
         <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Localization, which is a Quality, cannot be in the domain of classifies, which is Concept.

owl/EASE-OBJ.owl
```diff
@@ -120,36 +352,204 @@
...
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#isShapeOf -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#isShapeOf">
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isQualityOf"/>
         <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Shape"/>
         <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <owl:propertyChainAxiom rdf:parseType="Collection">
```
mpomarlan: Shape, which is a Quality, cannot be in the domain of classifies, which is Concept.

owl/EASE-OBJ.owl
```diff
@@ -438,891 +895,2301 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Region"/>
         <rdfs:subClassOf>
             <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFrictionValue"/>
-                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#double"/>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isRegionFor"/>
+                <owl:allValuesFrom>
+                    <owl:Restriction>
+                        <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isClassifiedBy"/>
```
mpomarlan: This restriction is impossible to fulfill. The range of isClassifiedBy is Concept, whereas Color is a Quality which is disjoint from Concept.

owl/EASE-OBJ.owl
```diff
@@ -438,891 +895,2301 @@
...
         <rdfs:subClassOf>
             <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isQualityOf"/>
-                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
-                <owl:onClass rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isDispositionOf"/>
+                <owl:allValuesFrom>
+                    <owl:Restriction>
+                        <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasQuality"/>
+                        <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Capacity"/>
```
mpomarlan: Nice!

mpomarlan:
hasColor, hasLocalization, hasShape and their inverses seem to be defined in an inconsistent manner to the DUL foundation at the moment. Let's fix that and this can be approved.

mpomarlan:
Further reasoner warnings:

hasColor, hasLocalization, hasShape and their inverses make the property hierarchy non-regular. This is because of the property chain definitions associated with them.

mpomarlan:
Warning: there are two "contains" properties, one in EASE-OBJ, one in EASE, and they are defined as subproperties of hasLocation and co-occurs respectively. Perhaps they can be merged into the same property.

mpomarlan:
Reasoner warning:

EASE-OBJ#isDescribedBy, affordsTask, affordsBearer, affordsTrigger are not simple properties (they are defined via property chain) therefore they cannot appear in "exactly 1" restrictions.

Note that there is an isDescribedBy property in DUL already, which complicates tracking down why EASE-OBJ#isDescribedBy is non-simple.

mpomarlan:
Reasoner warning: Color is subclass of Nothing.

This is because Color hasRegion some ColorRegion,
ColorRegion isRegionFor only (isClassifiedBy some Color)
range of isClassifiedBy is Concept
Concept is Object
Color is subclass of Extrinsic subclass of Quality
Object and Quality are disjoint.

owl/EASE-OBJ.owl
daniel86: that was an error

daniel86:
> hasColor, hasLocalization, hasShape and their inverses seem to be defined in an inconsistent manner to the DUL foundation at the moment. Let's fix that and this can be approved.

this has been fixed

daniel86:
> Reasoner warning:
> 
> EASE-OBJ#isDescribedBy, affordsTask, affordsBearer, affordsTrigger are not simple properties (they are defined via property chain) therefore they cannot appear in "exactly 1" restrictions.
> 
> Note that there is an isDescribedBy property in DUL already, which complicates tracking down why EASE-OBJ#isDescribedBy is non-simple.

EASE-OBJ#isDescribedBy does not seem to exist. there is a isQualityDescribedBy, however.

daniel86:
> Reasoner warning: Color is subclass of Nothing.
> 
> This is because Color hasRegion some ColorRegion,
> ColorRegion isRegionFor only (isClassifiedBy some Color)
> range of isClassifiedBy is Concept
> Concept is Object
> Color is subclass of Extrinsic subclass of Quality
> Object and Quality are disjoint.

should be fixed now

daniel86:
> Reasoner warning:
> 
> EASE-OBJ#isDescribedBy, affordsTask, affordsBearer, affordsTrigger are not simple properties (they are defined via property chain) therefore they cannot appear in "exactly 1" restrictions.
> 
> Note that there is an isDescribedBy property in DUL already, which complicates tracking down why EASE-OBJ#isDescribedBy is non-simple.

could be fixed. Disposition now does not use *affords* relations anymore together with cardinality restrictions. I hope this was the only place where they have been used.

daniel86:
> Warning: there are two "contains" properties, one in EASE-OBJ, one in EASE, and they are defined as subproperties of hasLocation and co-occurs respectively. Perhaps they can be merged into the same property.

this is a bit more difficult. it seems the contains in EASE does not imply some space, but is also used for intervals. I simply renamed the one in EASE-OBJ for now to *containsPhysically* and leave it to a later point to harmonize both

mpomarlan:
Ok I think I figured out why the reasoner doesn't like isDescribedBy in qualified cardinality restrictions.

isDescribedBy is no longer a simple property, because it is the inverse of describes, which has a subproperty describesAlteredQuality, which is defined via a chain and is non-simple. 

As per the OWL structural specification, this makes any object property reachable via superproperty or inverse property relations to be non-simple as well.

Two obvious ways to proceed:

-- alter the axioms that use isDescribedBy: so a Disposition would be described by some Affordance, rather than exactly 1 Affordance OR
-- alter the property tree to either make describesAlteredQuality no longer a subproperty of describes, or no longer a property chain+rewrite the axioms for alteration to not use the chain.

What do you think?

mpomarlan:
Had a look at the latest changes btw, this looks good! Just a couple reasoner annoyances to take care of and I think we're set.

daniel86:
> Ok I think I figured out why the reasoner doesn't like isDescribedBy in qualified cardinality restrictions.
> 
> isDescribedBy is no longer a simple property, because it is the inverse of describes, which has a subproperty describesAlteredQuality, which is defined via a chain and is non-simple.
> 
> As per the OWL structural specification, this makes any object property reachable via superproperty or inverse property relations to be non-simple as well.
> 
> Two obvious ways to proceed:
> 
> -- alter the axioms that use isDescribedBy: so a Disposition would be described by some Affordance, rather than exactly 1 Affordance OR
> -- alter the property tree to either make describesAlteredQuality no longer a subproperty of describes, or no longer a property chain+rewrite the axioms for alteration to not use the chain.
> 
> What do you think?

the first option could create problems in the future, when people use this property in cardinality restrictions. the second option would not imply describes anymore, which is the whole point of having this chain.
So both options seem to be not optimal.

It might be needed to remove describesAlteredQuality entirely, and to have larger axioms instead.

daniel86:
describesAlteredQuality has been removed

daniel86:
@mpomarlan please have a look again, the latest commit just removes this property.

daniel86:
> Yep, this seems to be ok! Thanks for the great work getting this monster of a pull request done!

:D 

thanks for your careful review, much appreciated!

Issue #35, opened 2019-10-28
The ontology can't be reasoned with


mpomarlan:
At least, not when using DL reasoners such as Pellet, Hermit, or basically anything else that Protege comes with.

Presumably some highly non-standard DL reasoner might handle it, or maybe a full fledged first-order theorem prover, but this somewhat defeats the purpose of having an ontology in OWL. It's supposed to be accessible to standard DL techniques.

The problem comes from temporal properties such as after, before, startedBy, finishedBy-- any combo of Transient, Asymmetric, Irreflexive is highly suspect, and if these occur in property chains as well then DL reasoners can't handle them.

Suggested fix: isolate the problem object properties into another module, on which hopefully we will almost never depend.

porzel:
Sounds like as plan.

mpomarlan:
Fixed by #37

PR #37, opened 2019-11-08
Moved full Allen axiomatization to optional module.

mpomarlan:
Fixes issue #35 

Changes: 

I. the following properties are now only declared transitive in EASE.owl:

startedBy, finishedBy, starts, finishes, before, during, contains, after

Rationale: [OWL restrictions on Axiom Closure](https://www.w3.org/TR/owl2-syntax/#The_Restrictions_on_the_Axiom_Closure). In brief, transitive properties are composite, which means non-simple, and only simple properties can be declared asymmetric or irreflexive.

II. the following properties are now declared asymmetric and irreflexive in EASE-Allen.owl:

startedBy, finishedBy, starts, finishes, before, during, contains, after

Rationale: to capture the Allen semantics. Note a change in "finishes". Originally, this was axiomatized in EASE.owl as transitive and irreflexive.

Now it is axiomatized in EASE.owl + EASE-Allen.owl as transitive, antisymmetric, irreflexive. The reason for the extra "antisymmetric" tag is because it seems to capture something about the "finishes" relation, as it is described in a comment in EASE.owl: if A finishes B, then A ends exactly where B ends, but B starts strictly before A. This implies if A finishes B, then it cannot be that B finishes A.

III. simultaneous is no longer reflexive

Rationale: reflexive properties hold for EVERY individual. However, simultaneous is a subproperty of co-occurs, which has Domain and Range Event. Note that asserting Domain and Range for simultaneous to be Entity has no effect:

Event subclassof Entity
co-occurs some Thing subclassof Event
simultaneous subpropertyof co-occurs
simultaneous some Thing subclassof Entity

is the same as
simultaneous some Thing subclassof Event

Further, Event hasParticipant some Object, meaning Object individuals must exist for Event individuals to exist. But Object and Event are disjoint. However, if simultaneous is reflexive, then an Object must be simultaneous to itself. But for the simultaneous property to apply, the Object must be an Event, which contradicts the disjointness axiom for the two concepts.


daniel86:
@mpomarlan but I am not sure about the domain/range change of simultaneous. I use this more general, also between individual tasks in workflows.

daniel86:
But it was not ok before, with 'co occurs' already having Event as domain and range. And I guess simultaneous also implies its an event. I would rather like to have it like for *overlaps* which has Entity as range and domain, but I couldn't come up with appropriate property names

PR #39, opened 2019-11-26
contact / connection / support / ...

daniel86:
- fixed that some terms are defined multiple times in different ways
- *Support* is not a disposition, it is a force event type! The disposition to be used as a placement surface I called *Deposition* instead
- *Support* is not a state  because *Support* is a process. I removed all references to *Support* in EASE-STATE.
- EASE-STATE defined some roles that were also defined in EASE-OBJ, I moved all role definitions from EASE-STATE to EASE-OBJ
- fixed an issue with *contains* relation being defined only for events before, now there are *containsEvent* and *containsObject* sub-relations
- introduced some *Contact* sub-classes: *Support*, *Collision*, *Fixation*, *Rubbing*, *Tapping*


owl/EASE-ACT.owl
```diff
@@ -9,9 +9,10 @@
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
     <owl:Ontology rdf:about="http://www.ease-crc.org/ont/EASE-ACT.owl">
-        <owl:imports rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl"/>
+        <owl:imports rdf:resource="package://ease_ontology/owl/EASE-PROC.owl"/>
```
daniel86: Reverted back to package url's to avoid issues with protege

owl/EASE.owl
```diff
@@ -192,14 +192,26 @@ Sub-properties are used to distinct between different cases of event endpoint re
...
+        <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
+    </owl:ObjectProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE.owl#containsEvent -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#containsEvent">
         <rdfs:subPropertyOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#coOccurs"/>
+        <rdfs:subPropertyOf rdf:resource="http://www.ease-crc.org/ont/EASE.owl#contains"/>
```
daniel86: need to have two super-properties

owl/EASE.owl
```diff
@@ -1131,27 +1112,6 @@ Such an event does NOT fall under the definition of Accident here. An example of
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE.owl#SocialState -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE.owl#SocialState">
```
daniel86: moved to EASE-STATE

owl/EASE-STATE.owl
```diff
@@ -73,41 +74,27 @@
...
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#UsuallyAgentive"/>
-    </owl:Class>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Container"/>
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#AssemblyConnection -->
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#LinkedObject -->
 
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#AssemblyConnection">
```
daniel86: for now we only have the Linkage* concepts and do not refer to *assembly*

owl/EASE-PROC.owl
```diff
@@ -149,95 +217,6 @@ As such, this concept can be applied only to motions involving some PhysicalAgen
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-PROC.owl#HeadTurning -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#HeadTurning">
```
daniel86: these were all just moved by protege. @mpomarlan please use protege to save the files to avoid this

owl/EASE-PROC.owl
```diff
@@ -149,95 +182,6 @@ As such, this concept can be applied only to motions involving some PhysicalAgen
...
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl#PosturalMoving"/>
-        <EASE:ELANName>move-body-turn</EASE:ELANName>
-        <EASE:ELANUsageGuideline>move own body, turn</EASE:ELANUsageGuideline>
-    </owl:Class>
-    
-
-
-    <!-- http://www.ease-crc.org/ont/EASE-PROC.owl#ArtifactContact -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#ArtifactContact">
```
mpomarlan: This should stay.

owl/EASE-PROC.owl
```diff
@@ -356,32 +309,6 @@ Note, most of the physical Entities we are concerned with are actually arrangeme
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-PROC.owl#EffectorContact -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#EffectorContact">
```
mpomarlan: This should stay.

owl/EASE-OBJ.owl
```diff
@@ -1160,44 +1180,14 @@
...
-                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#ConnectedObject"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#LinkedObject"/>
             </owl:Restriction>
         </rdfs:subClassOf>
         <rdfs:comment>The disposition of an object (the tool) to change the compositional structure of others.</rdfs:comment>
     </owl:Class>
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#ConnectedObject -->
```
mpomarlan: This shouldn't go away.

owl/EASE-OBJ.owl
```diff
@@ -1160,44 +1180,14 @@
...
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#ConnectedObject">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
-        <rdfs:comment>An object that is combined with another object.</rdfs:comment>
-    </owl:Class>
-    
-
-
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Connectivity -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Connectivity">
```
mpomarlan: This shouldn't go away.

owl/EASE-OBJ.owl
```diff
@@ -2457,45 +2480,6 @@
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Support -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Support">
```
mpomarlan: There should be a disposition for the ability to support something however. This needs more discussion before it can be committed.

mpomarlan:
I would suggest two things for future pull requests.

A) They should be smaller. For quality control reasons, a pull request should ideally address a well specified issue rather than being a kitchen sink in which to throw any kind of change.

B) Change is good, but lets be a little political about it. Refactoring of concepts is fine, removing concepts however should be done very conservatively. 

Ie., there should be more discussions about the approach to keep various branches of the ontology up to date with the latest modelling decisions, and these discussions should result in a documented plan, rather than just commit and hope the reviewer can understand why something happened.

owl/EASE-STATE.owl
daniel86: As pointed out in our meeting, the concept would move to either knowrob.owl or knowrob_assembly.owl. The rational is that assembly is not an everyday activity.

owl/EASE-STATE.owl
```diff
@@ -73,41 +74,27 @@
...
-                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#AssemblyConnectionConfiguration"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
-        <rdfs:comment>Classifies States in which two objects are in a rigid connection, such that the movement of one determines the movement of the other.</rdfs:comment>
-    </owl:Class>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#LinkedObject"/>
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#AssemblyConnectionConfiguration -->
```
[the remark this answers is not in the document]
daniel86: Same as above, the assembly concepts moved to another ontology. They were not erased.

owl/EASE-PROC.owl
```diff
@@ -234,12 +269,6 @@ In some sense any process that results in entities being created or destroyed mi
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#Creation">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl#ProcessType"/>
-        <rdfs:subClassOf>
```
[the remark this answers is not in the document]
daniel86: no they were invalid actually. Because they were rather supposed to link event and participant but we are here axiomatizing a *ProcessType* but not a *Process*.

owl/EASE-PROC.owl
```diff
@@ -273,12 +302,6 @@ A soft slab of clay deforming under its own weight on Earth would still count as
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#Destruction">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl#ProcessType"/>
-        <rdfs:subClassOf>
```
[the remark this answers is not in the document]
daniel86: same as above

owl/EASE-PROC.owl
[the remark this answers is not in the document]
daniel86: I assume someone has overseen this during refactoring before.

owl/EASE-OBJ.owl
```diff
@@ -1191,44 +1330,12 @@
...
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Container">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Restrictor"/>
         <rdfs:comment>An object used to contain others.</rdfs:comment>
     </owl:Class>
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Containment -->
```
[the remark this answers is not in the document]
daniel86: As explained in our meeting, the taxonomy below *Containment* was terrible. Containment only having one sub-class which is Enclosing which in turn has only one sub-class Insertion. For now I was only considering container where the robot can insert stuff, so i removed Containment and Enclosing in favor of Insertion. I do not state that this is the correct way a taxonomy should look like. But I perceive it as less of a hack for now, as it was not clear before where the distinctions are. As Robert pointed out, we might get a proper taxonomy based on image schema or so.

owl/EASE-OBJ.owl
```diff
@@ -1579,30 +1686,6 @@
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#EnclosedObject -->
```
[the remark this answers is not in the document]
daniel86: same as above

owl/EASE-OBJ.owl
```diff
@@ -1579,30 +1686,6 @@
...
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#EnclosedObject">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#IncludedObject"/>
-        <rdfs:comment>An object included within the spatial boundaries of another object.</rdfs:comment>
-    </owl:Class>
-    
-
-
-    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Enclosing -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Enclosing">
```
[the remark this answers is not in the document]
daniel86: same as above

owl/EASE-OBJ.owl
daniel86: There is. I called it *Deposition* as in something to deposit stuff

owl/EASE-OBJ.owl
```diff
@@ -2760,12 +2848,6 @@ It is also possible that a Transient transitionsBack to the initial object. An e
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE.owl#DesignedContainer">
         <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#DesignedArtifact"/>
-        <rdfs:subClassOf>
```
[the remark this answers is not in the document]
daniel86: Well because the Containment disposition got removed, and I was not sure if being a designed container actually implies that you can insert something.
If we want to assume that, then I would re-add the axiom for the Insertion disposition for now.

owl/EASE-OBJ.owl
daniel86: We have *LinkedObject* now. hence it did not went away but got refactored

owl/EASE-OBJ.owl
daniel86: It doesn't. We have *Linakge* instead.

owl/EASE-PROC.owl
daniel86: Why? It seems like a silly concept. It was only added for the very first version of the parser. I do not think we still need this.

owl/EASE-PROC.owl
daniel86: Why? I do not think this is very useful.

owl/EASE.owl
```diff
@@ -267,24 +267,6 @@ Sub-properties are used to distinct between different cases of event endpoint re
     
 
 
-    <!-- http://www.ease-crc.org/ont/EASE.owl#hasCreated -->
-
-    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#hasCreated">
```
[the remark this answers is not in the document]
daniel86: it was *invalid* as it is defined as sub-property of *isParticipantIn* -- you can re-add it if you like. But do not complain that I remove invalid axioms.

owl/EASE.owl
```diff
@@ -267,24 +267,6 @@ Sub-properties are used to distinct between different cases of event endpoint re
...
-        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isParticipantIn"/>
-        <rdfs:label>has created</rdfs:label>
-    </owl:ObjectProperty>
-    
-
-
-    <!-- http://www.ease-crc.org/ont/EASE.owl#hasDestroyed -->
-
-    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE.owl#hasDestroyed">
```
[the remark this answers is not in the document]
daniel86: same as above. It was *invalid*

daniel86:
> This pull request is way too aggressive in removing things that really shouldn't be removed.

Well the same I would state about your harsh review. Most stuff got refactored and not removed. Most changes were wrt. dispositions which are super new, where we only had a preliminary version with a bit unclean taxonomy below disposition. And the other branch of changes was about fixing issues that are currently existing in the repository.

daniel86:
> I would suggest two things for future pull requests.
> 
> A) They should be smaller. For quality control reasons, a pull request should ideally address a well specified issue rather than being a kitchen sink in which to throw any kind of change.
> 
> B) Change is good, but lets be a little political about it. Refactoring of concepts is fine, removing concepts however should be done very conservatively.
> 
> Ie., there should be more discussions about the approach to keep various branches of the ontology up to date with the latest modelling decisions, and these discussions should result in a documented plan, rather than just commit and hope the reviewer can understand why something happened.

sure, small pull requests are much nicer. However, I change the ontology for the purpose to use it for something in KnowRob. Hence, when I work on it then to realize a feature in KnowRob, and the changeset for the ontology will be accordingly. Often I could try to split this up into multiple ones, but this is often not so easy, especially when you do not have a concrete plan what needs to be done when you start working on it.

owl/EASE-OBJ.owl
daniel86: There is, it is called *Deposition*

daniel86:
So as far as I can see the following points still need discussion.

1. Do we need the concepts *ArtifactContact* etc. that classify *Contact* according to who is the participant? My proposal is rather to classify contact according to the *force interaction* of objects being in contact.
2. Is the shallow disposition taxonomy as it is ok for now? If not we need to figure out the relation between *Containment* and *Insertion*. I was not sure and wanted to delay this decision.
3. Is it ok to move the assembly concepts into KnowRob ontologies? In case not, we need to figure out the distinction between e.g. a regular *Connection* and an *AssemblyConnection*, I think this is not clear enough yet.
4. And the big one: Is support a State, is it a Process, or is it both? Is support a particular type of *Contact* or what is the relation between contact and support?

@mpomarlan anything I forgot here?

daniel86:
About *4*, our notion of state does make it a bit difficult to deal with *Contact*. As some contacts are no "stable" configurations, while other contacts are "stable". However, I think it makes sense to classify contacts according to their "force dynamic profile", i.e. how objects interact with each other. In that sense, e.g., *Support* and *Collision* would be clearly sub-concepts of *Contact*, I think. Generally speaking, the ontological category of *Support* and *Collision* seems very close as both are about objects being in contact and how they interact.

When we insist on processes implying *change* and state implying a *stable configuration*, I think the ontological category of *Force event type* is neither of them, or at least it is not obvious to which it might belong.

daniel86:
I will try to do s series of pull requests over the weekend. Then we can have a more focused discussion, and it will hopefully be easier to grasp the gist of the request.

PR #40, opened 2019-11-29
Added script to invoke hermit and produce warnings when some concept …

mpomarlan:
…is subsumed by Nothing.

To call, use something like

python3 hermit_test.py --reasoner /usr/local/HermiT/HermiT.jar EASE-uglyfied.owl

Outputs to stdout a list of items such as

WARNING: <http://example.org/owl/musteronto#Example> is subclass of Nothing

If everything is ok there should be no stdout output.

If the reasoner can't start at all, e.g. because the ontology is undecidable, then an Exception message is printed to that stdout output.

These texts could be piped into a file or wherever some other continuous integration script can make use of them.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -23,3 +23,16 @@ Dependencies:
 * ROS
 * rosprolog
 
+**Testing consistency**
+
+Run the hermit.py script to verify an OWL file with the HermiT reasoner. Verification here means checking whether some concepts are subsumed by owl:Nothing, ie. they are unsatisfiable.
+
+Syntax to run the script is:
+
+./hermit_test.py [options] inputfile
+
+where inputfile is a path to an OWL file. The following options are available:
+* -r, --reasoner: path to HermiT.jar, the executable for the HermiT reasoner. If not provided, it is assumed the jar is in the same folder as where the script is called from
+* -o, --output: path to a file where to store the results of the HermiT reasoner. If not provided, results are not stored
+
+Classification and subsumption reasoning will be performed on the input owl file. In case of non-satisfiable concepts, messages will be produced on stdout (ie. the console) by the script. A successful run of the script, that is, if no errors happen and no non-satisfiable concepts are found, produces no console output.
--- a/scripts/hermit_test.py
+++ b/scripts/hermit_test.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 import os
 import sys
 from optparse import OptionParser
--- a/scripts/hermit_test.py
+++ b/scripts/hermit_test.py
@@ -1,0 +1,64 @@
+import os
+import sys
+from optparse import OptionParser
+
+import subprocess
+import re
+
+usage = "usage: %prog [options] inputfile"
+parser = OptionParser(usage)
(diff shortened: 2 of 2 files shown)


shoeffner:
I tried it and interestingly this one has hickups when using the package:// URLs (here testing EASE-UGLY):
```
ERROR: reasoner failed to complete classification query
Feb 14, 2020 11:58:00 AM uk.ac.manchester.cs.owl.owlapi.ParsableOWLOntologyFactory canLoad
INFO: Malformed URL: unknown protocol: package
```

When I run HermiT from Protégé it works:
```
   INFO  11:55:52  ------------------------------- Running Reasoner -------------------------------
   INFO  11:55:52  Pre-computing inferences:
   INFO  11:55:52      - class hierarchy
   INFO  11:55:52      - object property hierarchy
   INFO  11:55:52      - data property hierarchy
   INFO  11:55:52      - class assertions
   INFO  11:55:52      - object property assertions
   INFO  11:55:52      - same individuals
   INFO  11:57:41  Ontologies processed in 109354 ms by HermiT
```
It has multiple issues, mostly evolving around, as far as I can see, `Social object` and `Event`.

mpomarlan:
Interesting: what package:// path is the EASE-UGLY.owl using?

mpomarlan:
Also, what issues are you seeing about Social object or Event?

shoeffner:
```
<owl:Ontology rdf:about="http://www.ease-crc.org/ont/EASE-UGLY.owl">
  <owl:imports rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl"/>
  <owl:imports rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/IOLite.owl"/>
  <owl:imports rdf:resource="package://ease_ontology/owl/EASE-PROC.owl"/>
  <owl:imports rdf:resource="package://ease_ontology/owl/EASE-STATE.owl"/>
  <owl:imports rdf:resource="package://ease_ontology/owl/EASE.owl"/>
</owl:Ontology>
```
These are the imports it uses.

All detected issues are:
image

And, just as an example, here are the justifications for `Contact`:

image

If I understand it correctly, then the problem is that `Contact` eventually inherits from `Object` via social object, which is disjoint with `Event`, but `Contact` is an `Event`.

mpomarlan:
Your understanding of the explanation seems correct. 

Hmm, interesting. I hoped that EASE-UGLY would only import DUL (which it can do via online URLs, which HermiT can handle). Fair point, I should have documented this. A small change to the README.md will be incoming.

The unsatisfiable classes are of course concerning-- and of course the reason why we would want to automate calling some reasoner for continuous integration and testing :)

 I'll open an issue about these unsatisfiable classes on Monday.

daniel86:
> 
> Hmm, interesting. I hoped that EASE-UGLY would only import DUL (which it can do via online URLs, which HermiT can handle). Fair point, I should have documented this. A small change to the README.md will be incoming.
> 
Yes me too, it is trivial to just ignore these imports in the Prolog code. I can change that until Monday.

shoeffner:
I can take care of the the automated testing next week.

PR #44, opened 2019-12-06
Refactor

daniel86:
- moved state concepts from EASE.owl to EASE-STATE.owl
- moved roles from EASE-STATE.owl to EASE-OBJ.owl and aligned a bit (I have removed the concepts *UsuallyAgentive* & *UsuallyThematic* as they seem to not fit well with the current role taxonomy in EASE-OBJ
- I removed for now invalid properties *hasCreated* and *hasDestroyed*, sorry that this is part of this pull request but it allowed me to do a cherry-pick. Let's re-add them in another pull request.

### DIFF
--- a/owl/EASE-OBJ.owl
+++ b/owl/EASE-OBJ.owl
@@ -854,6 +854,14 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Accessor -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Accessor">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+    </owl:Class>
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Affordance -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Affordance">
@@ -949,7 +957,7 @@
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Barrier -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Barrier">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Restrictor"/>
         <rdfs:comment>An object used to prevent others from entering or leaving a restricted space or group.</rdfs:comment>
     </owl:Class>
     
@@ -1201,7 +1209,7 @@
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Container -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Container">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Restrictor"/>
         <rdfs:comment>An object used to contain others.</rdfs:comment>
     </owl:Class>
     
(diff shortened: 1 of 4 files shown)


owl/EASE-OBJ.owl
```diff
@@ -1895,6 +1903,14 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Item -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Item">
```
daniel86: not sure if Item needs to stay, is it more specific then *Patient* ?

daniel86:
@mpomarlan the pull request has no conflicts anymore

owl/EASE-OBJ.owl
mpomarlan: I think it may have been intended as a more specific Patient, but it would need refactoring in any case since "Item" is too generic a name.

The "Item" comes from some FrameNET frames, where the thing being acted on is not any Patient (which can be animate or a person), but specifically an inanimate object.

owl/EASE-PROC.owl
```diff
@@ -234,12 +260,6 @@ In some sense any process that results in entities being created or destroyed mi
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#Creation">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl#ProcessType"/>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE.owl#hasCreated"/>
-                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
```
mpomarlan: It would be nice to keep the axiomatization I think, of course with the necessary correction.

So,

Creation subclassOf classifies only (hasCreated some PhysicalObject)

owl/EASE-PROC.owl
```diff
@@ -273,12 +293,6 @@ A soft slab of clay deforming under its own weight on Earth would still count as
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-PROC.owl#Destruction">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl#ProcessType"/>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE.owl#hasDestroyed"/>
```
mpomarlan: Similarly,

Destruction subclassOf classifies only (hasDestroyed some PhysicalObject)

daniel86: > The hasCreated, hasDestroyed axioms in the definition of process types Creation and Destruction should be corrected and kept I think. We may need to reason with entity creation or destruction in the cooking domain.

The thing is that then we would need to add a property for *all* the roles. As there is a role *DestroyedObject*, I think. I would like to keep redundancy low, and roles being concepts rather then properties.

Unfortunately, there is no fitting property for linking anything other then *Task* to roles (using isTaskOf). But I would rather vote for at some point start to axiomatize *EventTypes* with their roles and parameters -- this is not done yet at all.

daniel86: there is also a ticket about adding role/parameter axioms: https://github.com/ease-crc/ease_ontology/issues/24

mpomarlan: Ok, then how about

Creation subclassof classifies only ('has participant' some (PhysicalObject and 'is classified by' some CreatedObject)

Destruction subclassof classifies only ('has participant' some (PhysicalObject and 'is classified by' some DestroyedObject)

daniel86: but wouldn't that mean that the object is *always* classified as *CreatedObject*? Also, since the *Creation* concept has the role *CreatedObject* (at least in the future at some point), I think this would again be redundant. It should be enough to state that *Creation* is a process of the role *CreatedObject* and that this role only classifies *PhysicalObject*

mpomarlan: To answer the question, no. Here's a model of the axiom above for Creation:

let C be a Creation individual, therefore a Process Type. It classifies P, an individual of type Process. P hasParticipant O1 and O2, both of which are individuals of type PhysicalObject. O1 is classified by CO, an individual of type CreatedObject. O2 is NOT classified by any role.

As to the future of Creation, why not make the future now :) The axioms I suggested do exactly the thing you require: they link Creation and CreatedObjects as classifiers for Processes and some participants in them.

daniel86: As of the future is now: this is not the scope of the pull request. The axiomatization of process types with tasks and roles probably should be done in a separate effort.

daniel86: > O1 is classified by CO, an individual of type CreatedObject.

"O1 is classified by CO, an individual of type CreatedObject." -- no. It is classified as such only in the scope of the occurrence of a *Creation* process.

mpomarlan: Ok. Then we'll look into this after the holidays.

PR #46, opened 2019-12-06
+Feature

daniel86:
This pull request defines a *Feature* concept. The definition is taken from DOLCE. A feature is some physical aspect such as a bump or whole that also has a location.
However, there is no *PhysicalEndurant* in DUL which is the super-class of *Feature* in DOLCE. Hence, I defined it as sub-class of *Object*.

In KnowRob, static transforms are generated for features of the object where the transform is always given relative to the host of the feature. In particular, this is already used to assert that shelves in our retail store have a feature on the footer which is used to detect the shelve.

### DIFF
--- a/owl/EASE-OBJ.owl
+++ b/owl/EASE-OBJ.owl
@@ -240,6 +240,18 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFeature -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFeature">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasConstituent"/>
+        <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isFeatureOf"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <rdfs:range rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Feature"/>
+        <rdfs:label>has feature</rdfs:label>
+    </owl:ObjectProperty>
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFrictionAttribute -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFrictionAttribute">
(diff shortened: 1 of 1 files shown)


owl/EASE-OBJ.owl
```diff
@@ -1599,6 +1631,31 @@
...
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#isFeatureOf"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onClass rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <owl:disjointWith rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Agent"/>
+        <owl:disjointWith rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
```
daniel86: I guess this is a bit questionable. Let's consider the sticker on the footer of our retail shelves. It seems that the sticker is both a physical object that is a component of the shelve, but also a feature that is useful to detect a shelve.

owl/EASE-OBJ.owl
```diff
@@ -240,6 +240,18 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFeature -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#hasFeature">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasConstituent"/>
```
mpomarlan: I would lean towards hasPart as a superproperty, but I guess the idea with hasPart is that it points to entities that can be taken out of the whole and have independent existence, while these features do not?

owl/EASE-OBJ.owl
```diff
@@ -1373,6 +1396,15 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#DependentPlace -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#DependentPlace">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Feature"/>
+        <rdfs:comment>A feature that is not part of its host, like a hole in a piece of cheese.</rdfs:comment>
```
mpomarlan: Oh I see, so it's not a part because it can't be taken out and have an independent existence. That's ok.

owl/EASE-OBJ.owl
mpomarlan: We may consider some refactoring at some point. I think your idea to have features as constituents, rather than parts, is because the features do not exist independently of the whole. (A hole in the cheese can't exist without the cheese).

Under this logic, the sticker is a part that can have independent existence.

Perhaps both parts and features can be information bearing, and the borne information aids detection.

owl/EASE-OBJ.owl
```diff
@@ -2251,6 +2308,15 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#RelevantPart -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#RelevantPart">
```
mpomarlan: I would suggest this be named a bit differently so that it can't be confused with a part, especially that we seem to go towards a nice distinction between parts and features.

daniel86:
@mpomarlan the pull request has no conflicts anymore.

What should I still change taking into account our discussion?

mpomarlan:
I think we in the end sort of settled on keeping RelevantPart named as it is, and blame the DOLCE people for being careless :p

The other thing would have been whether features are constituents, on which neither of us had strong feelings if I remember correctly. Robert made an interesting case that constituency is about a kind of parthood across different layers of abstraction, which might actually fit, so for now we can keep this. 

I think there's a danger that Feature may be used in weird ways in the future. I am not clear now, post our discussion with Robert, what the difference between a Feature and a PhysicalObject is anymore; in my intuitive understanding, a corner, hole, or scratch were features. This is different to the texture of an animal's fur, and is also different from whether there exists a sticker with a marker pattern on a shelf.

Nevertheless, Feature can, in somewhat overloaded fashion, be used to refer to all of these things:

-- parasitic entities (holes, corners)
-- qualities (e.g. fur texture)
-- information-bearing and/or (detection) task-relevant parts that are themselves physical objects/artifacts (e.g. the sticker)

This is not a problem with what's in this pull request though. Rather it's a problem of what's not there yet, but we can have another round of development/pull request for that later.

PR #51, opened 2019-12-15
GitHub actions: Compile NEEM Spec, Uglify ontology

shoeffner:
This is an attempt at building the documentation and the uglified ontology using GitHub Actions.
I discussed with @porzel how to actually build the latest ontology and it took me some time to get all ROS dependencies together – is there an easier way than pulling in knowrob and all the sub dependencies?

Either way, this PR comes with two workflows:
- Build NEEM Specification
- Uglify Ontology

The Build NEEM Specification workflow downloads the Springer template and builds the docs/NEEM-specification/main.tex using the Makefile. I changed it so that the result is called neems.pdf, but I could change it back. The resulting file is uploaded as a build artifact (You can check out the [result of this branch on my fork](https://github.com/shoeffner/ease_ontology/commit/be1e01fa4f99da763a444229371fc6fa44044102/checks?check_suite_id=358812930), in the top right there is the artifact download).

The Uglify Ontology workflow installs all required ROS dependencies (and maybe some more, but at least it builds now) and links the source code directory into the catkin workspace, such that the EASE Ontology can be uglified properly. The [resulting artifact](https://github.com/shoeffner/ease_ontology/commit/5a8542c5e52bf24ff28fdd092af91b901fe9ae1f/checks?check_suite_id=358824875) contains all owl files, including the EASE-UGLY.owl.

Additionally I added the badges to the README, because, well, badges.

I am looking forward to your feedback!


daniel86:
Cool stuff @shoeffner !

Have you tried to run *uglify_ease.pl* without ROS installed? I think it should work, there is no dependency in the code to anything declared in KnowRob as far as I can see.
Hence we could probably avoid to set-up the ROS environment in the github action.

daniel86:
> Cool stuff @shoeffner !
> 
> Have you tried to run _uglify_ease.pl_ without ROS installed? I think it should work, there is no dependency in the code to anything declared in KnowRob as far as I can see.
> Hence we could probably avoid to set-up the ROS environment in the github action.

Oh sorry, I was wrong. There is a predicate `ros_path` used in this source file which is declared in *rosprolog*. But I still would prefer to have a duplicate declaration in *ease_ontology* such that we can run the script without ROS

shoeffner:
Thanks, I will try to run it without ROS, I would love to have a leaner
solution than pulling in a 5 GB docker image.
I haven't thought about running prolog alone as it looked like a ros
package... But I will!

daniel86:
It might be possible to use [source_file/2](https://www.swi-prolog.org/pldoc/doc_for?object=source_file/2), and then just restrict all OWL files to be at *$file_path/../owl*

shoeffner:
Thanks, using source_file/2 I was able to to build it using swipl alone – the github action is now blazingly fast, as it can use the official [swipl image](https://hub.docker.com/_/swipl/) without any changes.

shoeffner:
In its current state, when decoupling the ontology from ROS, the imports are not working properly: they use ROS' `package://...` URLs.

daniel86:
> In its current state, when decoupling the ontology from ROS, the imports are not working properly: they use ROS' `package://...` URLs.

yes, but the only ontologies imported are from the *ease_ontology* package, and we can find out the path to this package without ROS, e.g. by checking the path to the script

shoeffner:
I changed the compilation process for the main.pdf back and use the default docker image now. Unfortunately, the smaller latex images I tried are usually even worse than the one I use.

So the main CI time for the neem spec building is the pulling of the proper latex container.
I also tried building my own container based off the ctanbase image, but it turned out that installing all required packages individually just increased the CI time even further.

Once small victory though: the template gets cached now, that saves about... 4-8 seconds or so. At least I learned something new.

shoeffner:
- I resolved the issue with the package:// imports and changed all imports to the proper ontologies. - I added EASE-PROC to uglify, it was included but missing there.
- I added an additional build step using owlready2 to convert the resulting EASE-UGLY.owl to an owl-XML rather than an RDF XML file, as Protégé was not properly importing the latter.

owl/EASE-ACT.owl
```diff
@@ -1,18 +1,18 @@
...
      xmlns:xml="http://www.w3.org/XML/1998/namespace"
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
-     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
+     xmlns:ease="http://www.ease-crc.org/ont/EASE.owl#"
+     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+     xmlns:ease-proc="http://www.ease-crc.org/ont/EASE-PROC.owl#">
     <owl:Ontology rdf:about="http://www.ease-crc.org/ont/EASE-ACT.owl">
-        <owl:imports rdf:resource="package://ease_ontology/owl/EASE-PROC.owl"/>
-        <owl:imports rdf:resource="package://ease_ontology/owl/EASE-STATE.owl"/>
+        <owl:imports rdf:resource="http://www.ease-crc.org/ont/EASE-PROC.owl"/>
```
daniel86: This is problematic, I am afraid.

First of all, the files are still not there, http://www.ease-crc.org/ont/EASE-PROC.owl gives 404. Second, ease-crc returns a HTML page. I experienced problems clients not detecting this and trying to parse the HTML as OWL.
I know, this should be fixed.

Anyway, I was thinking that this can be ignored, as only other EASE ontologies are imported that way. Hence they do not need to appear in the uglified version. Then you could leave this hack for now.

owl/EASE-PROC.owl
```diff
@@ -36,7 +36,7 @@
         <rdfs:comment>A relation between a description and a process type.</rdfs:comment>
         <rdfs:label>defines process</rdfs:label>
     </owl:ObjectProperty>
-    
+
```
daniel86: it might be that you have used a different Protege version for editing, which causes the diff to be bloated.

Potentially only way to avoid ping-pong'ing large diffs is to use exactly the same version. We haven't documented this, but it is certainly needed to avoid these formatting diffs

shoeffner: I cleaned up the diffs. It was automatic trailing whitespace removal from another editor, mostly.

owl/EASE-ACT.owl
shoeffner: The `package:` variant also fails to resolve and is confusing if one wants to use the ontology without a ROS stack.
With this URL, at least manual debugging will help figuring out that the files give 404s.

Nevertheless, I can also change the URLs to match a resource where the files are available, e.g. GitHub. I could also try and change this PR in a way that it automatically builds the gh-pages branch with all resulting ontologies and serves them as https://ease-crc.github.io/ease_ontology/owl/EASE.owl etc.

daniel86: Yes it is confusing. But we cannot break KnowRob. The package URLs must remain until we have a proper solution.

shoeffner:
The new commit changes all URLs to `https://ease-crc.github.io/ease_ontology/EASE[.*].xml` and adds an action which deploys the owl files accordingly to the gh-pages branch (see my [fork](https://github.com/shoeffner/ease_ontology/tree/gh-pages) for an example) if run on the master branch.

It requires an [`ACCESS_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) secret (repository rights seem to be enough), which you can add via Settings:
image
(see also https://github.com/JamesIves/github-pages-deploy-action for more information)

The results are then deployed to https://ease-crc.github.io/ease_ontology/EASE.owl etc.
As mentioned above, I deployed them on my fork, so you can find them e.g. at https://shoeffner.github.io/ease_ontology/EASE.owl.

daniel86:
> The new commit changes all URLs to `https://ease-crc.github.io/ease_ontology/EASE[.*].xml` and adds an action which deploys the owl files accordingly to the gh-pages branch (see my [fork](https://github.com/shoeffner/ease_ontology/tree/gh-pages) for an example) if run on the master branch.
> 
> It requires an [`ACCESS_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) secret (repository rights seem to be enough), which you can add via Settings:
> image
> (see also https://github.com/JamesIves/github-pages-deploy-action for more information)
> 
> The results are then deployed to https://ease-crc.github.io/ease_ontology/EASE.owl etc.
> As mentioned above, I deployed them on my fork, so you can find them e.g. at https://shoeffner.github.io/ease_ontology/EASE.owl.

yes, something like this makes sense. But KnowRob must be able to run offline. So it is not ok that KnowRob loads ontologies from http.

daniel86:
I would really prefer more atomic pull requests.
A pull request about automatically creating an artifact from files in the repo should only add this job, without touching any of the versioned files.
Then a second request can tackle the deployment of OWl files and so on. That would make it easier for me to accept this request.
For now I cannot merge as changing the IRIs of all the concepts breaks things -- also I do not see a reason why the uglifying action would require it.

shoeffner:
The uglifying action does not require changing all IRIs but the deployment step: because after deployment the IRIs should point to the correct locations, so it is necessary to do this for the proposed hosting solution.

But it is okay that this solution does not work, I hope a future attempt might resolve the knowrob compatibility issue.

daniel86:
Well the uglify action I would merge right away. are you planning to do another request just with the uglify action? Else I would manually copy some parts of your pull request into the source tree at some point (maybe this weekend in case I find the time)

shoeffner:
I have extracted only the build files etc. to #53, to have a clean and slightly more atomic PR.

daniel86:
thanks a lot!

PR #53, opened 2020-02-13
Adding workflow to uglify and build the docs.

shoeffner:
This commit uglifies the ontology without the need of ROS, converts it
to a proper owl file, and uploads the artifacts.
This commit also adds automatic building for the specification docs.
It updates the README to contain badges for both jobs.
Last, it adds an optional step which pushes the compiled ontologies onto
the gh-pages branch if a secrets.ACCESS_TOKEN is set and the build
happened on the master branch.

### DIFF
--- a/.github/workflows/ontology.yml
+++ b/.github/workflows/ontology.yml
@@ -1,0 +1,45 @@
+name: Uglify Ontology
+on:
+  push:
+    paths:
+      - '**'
+      - '!docs/**'
+
+jobs:
+  uglify_ontology:
+    name: Uglify Ontology
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+
+      - name: Uglify
+        uses: docker://swipl:8.0.3
+        with:
+          args: swipl -f prolog/uglify_ease.pl -t uglify_ease
+
+      - name: Convert
+        uses: ./.github/actions/convert
+        with:
+          args: python3 ./scripts/uglify_ease_as_owl.py
+
+      - name: Upload
+        uses: actions/upload-artifact@v1
+        with:
+          name: EASE Ontology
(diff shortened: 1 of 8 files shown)


scripts/uglify_ease_as_owl.py
```diff
@@ -0,0 +1,25 @@
...
+
+import owlready2
+
+
+def convert():
+    ease_ugly = Path(__file__).parent.parent / 'owl' / 'EASE-UGLY.owl'
+    if not ease_ugly.exists():
+        print('Generating rdf')
+        command = 'swipl -f prolog/uglify_ease.pl -t uglify_ease'.split()
+        proc = subprocess.run(command, cwd='data/ontologies/ease_ontology', capture_output=True)
```
daniel86: Is there any reason to still use the Prolog code at all? Couldn't we load all ontologies with owlready, and then export them?

It allows us to do transformation while loading.

.github/workflows/ontology.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Uglify
+        uses: docker://swipl:8.0.3
+        with:
+          args: swipl -f prolog/uglify_ease.pl -t uglify_ease
+
+      - name: Convert
+        uses: ./.github/actions/convert
+        with:
+          args: python3 ./scripts/uglify_ease_as_owl.py
```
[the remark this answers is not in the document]
shoeffner: In fact it does not seem to make much difference anymore, maybe I had issues with caching etc.?
It could be that the fix was really to change the package:// URLs and not converting it, so essentially we could remove this step.

scripts/uglify_ease_as_owl.py
shoeffner: With owlready2, it would write a very short owl file which only features a few import statements, while the prolog code copies many triples directly, I don't know which one would be a better solution.

PR #56, opened 2020-02-14
ignore all imports of EASE ontology modules in uglify script

daniel86:
This will cause EASE-UGLY to not have package imports any more.

### DIFF
--- a/prolog/uglify_ease.pl
+++ b/prolog/uglify_ease.pl
@@ -14,13 +14,21 @@ ease_assert(_Graph, rdf(_S,_P,O)) :-
 
 ease_assert(Graph, rdf(_S,P,O)) :-
   rdf_equal(P,owl:imports),!,
-  ( rdf_split_url('http://www.ease-crc.org/ont/',_,O) -> true ; (
-    ease_ugly_ontology(EASE_UGLY),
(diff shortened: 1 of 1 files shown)


daniel86:
@shoeffner I also tried to open generated EASE-UGLY in Protege and it did not complain

prolog/uglify_ease.pl
```diff
@@ -14,14 +14,22 @@
...
-    ease_ugly_ontology(EASE_UGLY),
-    rdf_assert(EASE_UGLY,P,O,Graph)
-  )).
+  ease_assert_import(Graph,O).
 
 ease_assert(Graph, rdf(S,P,O)) :-
   rdf_assert(S,P,O,Graph).
 
+%%
+ease_assert_import(Graph,Ontology) :-
```
shoeffner: `Graph` is now a Singleton variable and could thus be `_Graph` to denote it's essentially not used.

shoeffner:
Ah, I don't have write access, so my review does not count – @mpomarlan , maybe you can add one as well.

daniel86:
> Ah, I don't have write access, so my review does not count – @mpomarlan , maybe you can add one as well.

You should have write access now, use it wisely ;)

PR #62, opened 2020-02-29
Proctype fix

daniel86:
This pull request fixes some faulty *ProcessType* definitions using a new relation that is used to relate process types to roles.
Before they were axiomatized as if they were events, restricing their *participants*. But concepts may not have any participants.

depends on #61

### DIFF
--- a/owl/EASE-PROC.owl
+++ b/owl/EASE-PROC.owl
@@ -18,6 +18,29 @@
     <!-- 
     ///////////////////////////////////////////////////////////////////////////////////////
     //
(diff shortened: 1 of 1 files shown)


daniel86:
> Oops, git interface confusion on my part. So yes, the question is whether EASE-OBJ is a good import for EASE-PROC. Otherwise I think this looks good.

https://github.com/ease-crc/ease_ontology/pull/62/commits/9c784354d1566ff8353c904741f587d7a4196b5e adds the missing import

PR #63, opened 2020-03-03
+isGestalltOf -specific configuration concepts

daniel86:
This pull requests defines missing relations to associate *Gestallts* with *Roles* (similar to *hasTask* and *isTaskOf*).

Furthermore, I started using these relations for all *Gestallt* concepts. However, this makes the specific *Configuration* definitions obsolete (as they were only defining the roles). Hence, I removed them. Only the toplevel *Configuration* class remains as it was.

An argument for doing this is to keep it consistent with how we handle other types of events.
Another is that Configuration and Gestallt branches had redundant structure which makes modelling a bit more cumbersome (as you potentially need to do changes in two places)

### DIFF
--- a/owl/EASE-STATE.owl
+++ b/owl/EASE-STATE.owl
@@ -27,6 +27,31 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#hasGestallt -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#hasGestallt">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isRelatedToConcept"/>
+        <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#isGestalltOf"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:range rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#Gestallt"/>
+        <rdfs:comment>A relation between roles and gestallts, e.g. &apos;the chair is the supporter of the person sitting on it&apos;.</rdfs:comment>
+        <rdfs:label>has gestallt</rdfs:label>
+    </owl:ObjectProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#isGestalltOf -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#isGestalltOf">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isRelatedToConcept"/>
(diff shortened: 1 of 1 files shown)


mpomarlan:
I'm starting a review but I don't think I'll finish today.

owl/EASE-STATE.owl
```diff
@@ -27,6 +27,31 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#hasGestallt -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#hasGestallt">
```
mpomarlan: This might be a good time to rename Gestallt to StateType, and likewise for obj.prop. names.

Meaning, if I don't suggest any other changes, ignore this request; if I do request other changes, might as well put this one along too.

owl/EASE-STATE.owl
```diff
@@ -151,22 +183,13 @@ Several other Description subclasses may function as Configurations. For example
...
-
-
-    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#ContainmentConfiguration -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#ContainmentConfiguration">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#FunctionalControlConfiguration"/>
         <rdfs:subClassOf>
             <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#definesRole"/>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#isGestalltOf"/>
```
mpomarlan: Ah right, so since StateTypes/Gestallts are concepts, they should not use "defines" when linking to Roles (the way Situations or Descriptions might). Ok, makes sense.

owl/EASE-STATE.owl
```diff
@@ -219,27 +233,7 @@ Note that this State focuses on how the interaction of, usually non-agentive, ob
                 <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-STATE.owl#Gestallt"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-    </owl:Class>
-    
-
-
-    <!-- http://www.ease-crc.org/ont/EASE-STATE.owl#LinkageConfiguration -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-STATE.owl#LinkageConfiguration">
```
mpomarlan: Linkage is gone because we no longer put the assembly stuff in the EASE ontos, right?

mpomarlan:
Actually scratch that, I got through the changes. So basically we do away with a taxonomy of Configurations, and just include informations about Roles of participants via the hasGestallt/isGestalltOf properties.

Sounds good. We'll do renaming on a future occasion.

owl/EASE-STATE.owl
daniel86: true, I remember that this was suggested. However I like the term *Gestallt*. But it's true that meaning of StateType is much more obvious

shoeffner:
What does Gestallt mean? Should it be Gestalt?

PR #68, opened 2020-03-26
Planning and Taxis

daniel86:
- fixed faulty definition of Planning and Taxis
- added Taxis comment
- removed Taxis Locomotion axioms

The comment provides counter example of Taxis without Locomotion, the axioms were further inconsistent as they were (as they were referring to event_types).

### DIFF
--- a/owl/EASE-ACT.owl
+++ b/owl/EASE-ACT.owl
@@ -9,10 +9,10 @@
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
     <owl:Ontology rdf:about="http://www.ease-crc.org/ont/EASE-ACT.owl">
+        <owl:imports rdf:resource="package://ease_ontology/owl/EASE.owl"/>
         <owl:imports rdf:resource="package://ease_ontology/owl/EASE-PROC.owl"/>
         <owl:imports rdf:resource="package://ease_ontology/owl/EASE-STATE.owl"/>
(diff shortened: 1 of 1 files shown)


owl/EASE-ACT.owl
```diff
@@ -837,8 +837,13 @@ Note that buying an object is NOT PhysicalGetting. Buying, or ownership transfer
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-ACT.owl#Prospecting"/>
         <rdfs:subClassOf>
             <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasParticipant"/>
-                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Plan"/>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#isTaskOf"/>
```
mpomarlan: Can you explain this definition? If I understand it correctly, it is saying that Planning is a subclassof isTaskOf some (classifies some Plan)

daniel86: *Planning* is a sub-class of *Task*. So there are objects playing a role when the task is executed (*isTaskOf*), and one of these roles classifies the *Plan* (basically the *Outcome* or so)

daniel86: So before the statement was "Planning has some participant which is a plan", and this I have changed to "Planning is a task of some role that is taken by the plan"

PR #73, opened 2020-04-21
CRAM Logger related changes

seba90:
I added changes to the ontology so I am able to create robot NEEMs with CRAM.
I am aware that those changes are not conform with your ontology standard. 
If you have any question regarding the changes feel free to contact me. You are free to update the structure of the ontology without consulting me. However, if you need/want to change the URLs, please let me know, so I can incorporate those changes into my logging tool.


daniel86:
I think we can close this as the discussion was happening in the issue tracker anyway

PR #76, opened 2020-04-27
Geometry branch

AbhijitVyas:
Hi,

This is the first step in order to define some design concepts. Basically it is needed to relate axioms which describes Structural design.
I also have added additional "Size" as intrinsic quality.
Please take a look and see if it suits our needs.

Thanks and Best,
Abhijit

### DIFF
--- a/owl/EASE-OBJ.owl
+++ b/owl/EASE-OBJ.owl
@@ -1880,7 +1880,7 @@
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#GeometricalQuality">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
-        <rdfs:comment>An intrinsic quality that is an aspect of the spatial form of an object such as shape, size, sharpness etc</rdfs:comment>
+        <rdfs:comment>An intrinsic quality that is an aspect of the spatial form of an object such as it&apos;s shape, size, sharpness.</rdfs:comment>
     </owl:Class>
     
 
--- a/owl/EASE-OBJ.owl
+++ b/owl/EASE-OBJ.owl
@@ -1876,6 +1876,16 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
+        <rdfs:comment>Geometry is a branch of mathematics concerned with questions of shape, size, relative position of figures, and the properties of space.
+https://en.wikipedia.org/wiki/Geometry</rdfs:comment>
+    </owl:Class>
+    
+
+
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#GreenColor -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#GreenColor">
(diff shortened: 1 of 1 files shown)


owl/EASE-OBJ.owl
```diff
@@ -1876,6 +1876,16 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
+        <rdfs:comment>Geometry is a branch of mathematics concerned with questions of shape, size, relative position of figures, and the properties of space.
+https://en.wikipedia.org/wiki/Geometry</rdfs:comment>
```
daniel86, line 1884: We do not usually add links in the comment.
There are special annotation properties to refer to URLs that should be rather used then including a link in the comment text.

In this case, wikipedia has a citation for their definition. I would also prefer to refer to this.

In comments in DUL, there is often some information about different ontological perspectives on the term, sometimes with references to Literature.

owl/EASE-OBJ.owl
```diff
@@ -2564,7 +2574,9 @@
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Sharpness -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Sharpness">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry"/>
+        <rdfs:comment>The quality of having a thin edge or point that can cut something or make a hole in something.
+https://dictionary.cambridge.org/dictionary/english/sharpness</rdfs:comment>
```
daniel86, line 2579: same as above.

owl/EASE-OBJ.owl
```diff
@@ -2564,7 +2574,9 @@
     <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Sharpness -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Sharpness">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry"/>
+        <rdfs:comment>The quality of having a thin edge or point that can cut something or make a hole in something.
```
daniel86, line 2578: Seems like a good fir for the category :) thanks for adding the comment.
But you can also have a "sharp thought" or see a "sharp movie".
What about these cases? Is e.g. the sharpness of a movie also covered in this category?

owl/EASE-OBJ.owl
```diff
@@ -2595,6 +2607,16 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Size -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Size">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Geometry"/>
+        <rdfs:comment>Size is the magnitude or dimensions of a thing. Size can be measured as length, width, height, diameter, perimeter, area, volume, or mass.
```
daniel86, line 2614: *dimension* (without a s)

owl/EASE-OBJ.owl
AbhijitVyas: Done. Now reference urls are mentioned with SeeAlso tag.

owl/EASE-OBJ.owl
AbhijitVyas: Done.

owl/EASE-OBJ.owl
AbhijitVyas: The comment related to sharpness is updated to only focus on the concept where it comes to sharp object. Rest of the concepts needs to be defined for future scenarios.

owl/EASE-OBJ.owl
```diff
@@ -1876,6 +1876,15 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#GeometricalQuality -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#GeometricalQuality">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Intrinsic"/>
+        <rdfs:comment>An intrinsic quality that is an aspect of the spatial form of an object such as shape, size, sharpness etc</rdfs:comment>
```
daniel86, line 1883: Other comments end with a dot.

daniel86: I think, when you start with "such as" no etc. is needed at the end of the sentence, or?

daniel86: Maybe better "such as *its* shape, size, or sharpness."?

AbhijitVyas: Done.

PR #90, opened 2020-07-09
Role taxonomy, removal of some duplicate concepts

### LINKED ISSUE #84: Name clash: LinguisticObject
Name conflict between EASE-SAY and DUL about LinguisticObject. Both owls define a concept of this name. The overlap is significant however, so perhaps we can just make it so the EASE-SAY uses the DUL concept directly.

mpomarlan:
Added role taxonomy for basic roles Agent, Patient, Source, Goal, Path etc.

Fixes #84 , #91

More is coming, specifically the other roles appearing in VerbNET and their selectional restrictions, but this PR is getting large already.


daniel86:
Is this also about the card **Role Taxonomy** in the *Projects* view of github?

If so, the workflow would be to turn this card into a ticket and then associate it to the pull request.

owl/EASE-SAY.owl
```diff
@@ -82,13 +98,68 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#CausalProcessRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Refers to the ExecutableSchema which is causing a motion.</rdfs:comment>
```
daniel86, line 111: The comment should state what it **is**  , and not what it refers to

owl/EASE-SAY.owl
```diff
@@ -113,6 +184,89 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: A functional relation describing that the location of some entity, the containee, is constrained to be within some region which is the interior of some other entity, the container.
```
daniel86, line 191: what is the GUM-4-space doing in the beginning?
You should do citations as you would do in scientific publications.

owl/EASE-SAY.owl
```diff
@@ -113,6 +184,89 @@
...
+
+This is also known as FunctionalControlInternal in GUM.</rdfs:comment>
```
daniel86, line 193: What is GUM? Better add a reference here if it is important what it is in GUM.

owl/EASE-SAY.owl
```diff
@@ -113,6 +184,89 @@
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#FunctionalSpatialSchemaTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: The binary functional relation holding between two spatial objects x and y, such that x controls y&apos;s position in space; if x moves, then y moves as well; x and y need not be in contact.</rdfs:comment>
```
daniel86, line 202: same as above with GUM-4-space in the beginning.

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#RelatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: The most general type of functional spatial relations.</rdfs:comment>
```
daniel86, line 343: same as above with GUM-4-space in the beginning.

owl/EASE-SAY.owl
```diff
@@ -211,20 +453,113 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#RelatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: A qualitative spatial relation indicating relative proximity, as expressed by the prepositions &apos;near&apos;, &apos;close to&apos;, &apos;next to&apos;, etc.
```
daniel86, line 558: same as above with GUM-4-space in the beginning.

owl/EASE-SAY.owl
```diff
@@ -310,6 +702,23 @@ Note that such theories are useful both when analyzing an actual linguistic prod
...
         <rdfs:comment>A transition between two states brought about by the Action of some Agent.</rdfs:comment>
     </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#SupportTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#SupportTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: The reified functional relation holding between two spatial objects x and y, such that x physically supports y in the presence of gravity; x and y need not be in contact. An example of such an expression is &quot;The bowl is on the table&quot;.
```
daniel86, line 712: same as above with GUM-4-space in the beginning.

owl/EASE-SAY.owl
daniel86, line 202: this does not seem like a relation, maybe a reification of one.
- Is this the case?
- If so, how do we refer to the formal counterpart?
- Is there a formal relation *controls* defined at the moment?

daniel86: have a look at the comment of CoverageTheory, I like it more how it starts

"A schematic theory of a functional relation .... "

daniel86, line 202: what about roles in this theory?

owl/EASE-SAY.owl
```diff
@@ -113,6 +184,89 @@
...
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#FunctionalSpatialSchemaTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: The binary functional relation holding between two spatial objects x and y, such that x controls y&apos;s position in space; if x moves, then y moves as well; x and y need not be in contact.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#CoverageTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#CoverageTheory">
```
daniel86, line 209: role axioms are missing

owl/EASE-SAY.owl
```diff
@@ -113,6 +184,89 @@
...
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: A functional relation describing that the location of some entity, the containee, is constrained to be within some region which is the interior of some other entity, the container.
+
+This is also known as FunctionalControlInternal in GUM.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory">
```
daniel86, line 200: role axioms are missing

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#EntityRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes an existing entity, included in ESTSchemas.</rdfs:comment>
```
daniel86, line 305: that comment seems strange to me.
Roles are used to classify objects, sure in a linguistic context they are also used to denote entities, but from a formal viewpoint they are concepts used to classify.
Also what is an ESTSchema?

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#EntityRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#EntityRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
```
daniel86, line 302: restriction does not seem to fit with the concept name "EntityRole" -- so it is only a "PhysicalObjectRole"?

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes an existing entity, included in ESTSchemas.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#SchematicTheory"/>
+        <rdfs:comment>Also known as &quot;executing schemas&quot; or &quot;x-schemas&quot;, these were defined by Bergen and Chang in their work &quot;Embodied Construction Grammar in Simulation-Based Language Understanding&quot; as:
```
daniel86, line 314: Best would be to have one explanatory sentence what this is, then make the references.

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes an existing entity, included in ESTSchemas.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory">
```
daniel86, line 312: role axioms missing

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">GUM-4-space: The most general type of functional spatial relations.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#GoalRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#GoalRole">
```
daniel86, line 350: "GoalRole" because "Goal" is already taken for plans?

owl/EASE-SAY.owl
```diff
@@ -123,6 +277,83 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes the destination of a route. Typically represented by a FunctionalSpatialSchema.</rdfs:comment>
```
daniel86, line 352: Why not name it "Destination" then?
"Goal" seems to be more broad.

owl/EASE-SAY.owl
```diff
@@ -171,6 +404,21 @@ Image schemas are considered as fundamental, pre-conceptual, building blocks of
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#InstrumentRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#InstrumentRole">
```
daniel86, line 409: What is the plan with conflicting roles in EASE-OBJ?
As Instrument is definitely defined there.
Why not merging with it right away?

owl/EASE-SAY.owl
```diff
@@ -171,6 +404,21 @@ Image schemas are considered as fundamental, pre-conceptual, building blocks of
...
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalArtifact"/>
```
daniel86, line 414: a role can exist without ever being manifested, so universal restriction should be used instead

owl/EASE-SAY.owl
daniel86, line 302: a role can exist without ever being manifested, so universal restriction should be used instead

owl/EASE-SAY.owl
```diff
@@ -183,23 +431,17 @@ Note that in a semantic sense such clauses always request information, but in a
...
-        <rdfs:label>Linguistic object</rdfs:label>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes the object with primary focal prominence in a spatial (-temporal) schema. Terminological variants: Figure (Talmy 1983), Trajector (Langacker 1986).</rdfs:comment>
```
daniel86, line 444: why the minus in (-temporal)?

owl/EASE-SAY.owl
```diff
@@ -183,23 +431,17 @@ Note that in a semantic sense such clauses always request information, but in a
...
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#LinguisticObject">
-        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationObject"/>
-        <rdfs:comment>An information bearing object that is used to analyze some linguistic production.</rdfs:comment>
-        <rdfs:label>Linguistic object</rdfs:label>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
```
daniel86, line 441: a role can exist without ever being manifested, so universal restriction should be used instead

owl/EASE-SAY.owl
```diff
@@ -211,20 +453,113 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#InstrumentRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#PatientRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory for how an agent can mix a fluid or particulate object.</rdfs:comment>
```
daniel86, line 486: what is the relationship here?

owl/EASE-SAY.owl
vanjaSophie: The naming of "Goal" is simply due to cognitive linguistics heritage, but we can of course also merge it to the existing concept.

owl/EASE-SAY.owl
vanjaSophie: Yes, I think we could definitely merge the corresponding roles from EASE-OBJ, e.g. Destination, Instrument, and maybe also Patient.

owl/EASE-SAY.owl
```diff
@@ -63,6 +64,21 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#AgentRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalAgent"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Semantic macrorole often employed in ExecutableSchemas that generalises across thematic relations, such as agent or experiencer, which themselves abstract away over verb-specific semantic roles.</rdfs:comment>
```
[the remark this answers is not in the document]
vanjaSophie: Maybe we are thinking about "thematic" in different contexts. ^^' [The one I meant was this linguistic one.](https://en.wikipedia.org/wiki/Thematic_relation)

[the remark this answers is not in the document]
daniel86: thanks for providing the link!
Can it be written a bit more clear for people without much NLP background?

owl/EASE-SAY.owl
daniel86: But for me it seems a destination is just one example of a goal. Another one would be to win a soccer tournament, or to achieve a degree, etc which would rather count as "destination" only metaphorically.

daniel86: FYI: the term "Goal" is used in DUL for plans, as in "plans are descriptions having an explicit goal". In that definition a "Goal" is a description of a (desired) situation.

daniel86:
> I stopped after 30 comments. Please consider my comments also for the parts I haven't directly addressed.
> 
> One bigger issue is with the relationships that correspond to a theory. As I think they also should exist formally (as an OWL property) and should be referred to explicitly by the theories (or at least in the comment).
> Some theories also do not provide information about roles while others come with such axioms.
> 
> Another concern is that a lot of roles are duplicates of roles in EASE-OBJ

@vanjaSophie please re-request a review when you think I should have a look again. You can also mark conversations as resolved if you are sure that the comment is taken care of.

owl/EASE-SAY.owl
```diff
@@ -82,13 +98,68 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#PatientRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasPart"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#SourcePathGoalTheory"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A has caused B to move C by having done D.
```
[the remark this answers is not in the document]
mpomarlan: Can you explain what the revision should accomplish?

owl/EASE-SAY.owl
mpomarlan: It inherits axioms from the parent. Basically, the same roles that are important for the superclass are important here as well and there is no need to have the same axioms twice.

owl/EASE-SAY.owl
mpomarlan: Nope, same as above, the role axioms come from the superclass.

owl/EASE-SAY.owl
mpomarlan: Role axioms come from the superclass.

owl/EASE-SAY.owl
mpomarlan: Not sure I follow. Is the exposition somehow unclear? We say where the notion comes from, ie. its original conceptualization that is relevant for us, and then modify that notion to suit SOMA needs.

owl/EASE-SAY.owl
mpomarlan: In this case I agree :) Note however that moving Agent and Patient role axioms here may leave other subclasses of this without axioms defining roles. As I argue above, this will actually be ok.

owl/EASE-SAY.owl
mpomarlan: Yes, and likewise for everything else. We would have naming conflicts for AgentRole/Agent otherwise etc.

owl/EASE-SAY.owl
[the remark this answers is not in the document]
daniel86: Well the comment should always answer the question "What is this concept" -- and this comment does not explicitely answer this question.

owl/EASE-SAY.owl
daniel86: no, what I meant is that in this comment, you first provide an aka sentence before stating what the concept is. My suggestion is to first say what this concept is and afterwards refer to x-schema and co.

owl/EASE-SAY.owl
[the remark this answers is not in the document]
daniel86: I think best practice if it is always possible to prepend "$CONCEPT is ..." on front of the comment

owl/EASE-SAY.owl
mpomarlan: But that's fairly common to do in dictionary definitions, e.g. the wikipedia article for first order logic (https://en.wikipedia.org/wiki/First-order_logic).

daniel86: well it is not critical, I agree. But if you look at comments in DUL.owl, you will see they almost all follow the same pattern. Anyway, it can stay like this

mpomarlan:
Also fixes #91

owl/EASE-SAY.owl
vanjaSophie: Ah, thanks for clarifying! In light of this definition, it makes sense to rename and merge `GoalRole` with the more primitive `Destination` concept.

mpomarlan:
Note: at the moment, there is no associated relation to SourcePathGoal.

Associating relations to functional control schemas is easy; they have two essential participants anyway, which makes the corresponding relation very easy to identify. This is not the case for SourcePathGoal, as there are 3 entities here, and at least (Source, Path) and (Path, Goal) look like they could be interesting relations. I left this to be discussed later.

owl/EASE-SAY.owl
```diff
@@ -63,6 +111,33 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#AgentRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:allValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalAgent"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A role often employed in ExecutableSchemas that generalises across thematic relations, such as agent or experiencer, which themselves abstract away from verb-specific roles.</rdfs:comment>
```
daniel86, line 136: I do not get the difference between "thematic relations" and "verb-specific roles".
Could you rephrase or elaborate more (e.g. provide an example that showcases difference)?

owl/EASE-SAY.owl
```diff
@@ -82,13 +157,69 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#CausalProcessRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#ExecutableSchematicTheory"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes some physical process that plays the role of causing a motion, e.g. motions, forces, etc.</rdfs:comment>
```
daniel86, line 170: I would suggest:

"A role that denotes physical processes causing a motion such as the force interaction of a tennis racket hitting a ball and setting it into motion"

owl/EASE-SAY.owl
```diff
@@ -63,6 +111,33 @@
...
+
+    <!-- http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#AgentRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#AgentRole">
```
daniel86, line 128: It is very common to speak of "agents" in other contexts, e.g. "cleaning agent".
Would they be included with this role definition?

owl/EASE-SAY.owl
daniel86: maybe just remove "which themselves abstract away from verb-specific roles"?

owl/EASE-SAY.owl
```diff
@@ -82,13 +157,69 @@
...
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
```
daniel86, line 188: so this is a many-to-many relation?
Or another reason to not use *exactly 1* here?

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#ControlTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A functional relation describing that the location of some entity, the containee, is constrained to be within some region which is the interior of some other entity, the container.
```
daniel86, line 251: Could you rephrase the beginning. It reads like "ContainmentTheory is a functional relation" -- this seems not right, or?

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#ContainmentTheory">
```
daniel86, line 249: I think role axioms are missing.
**containee** and **container** seem to be not defined in *ControlTheory*, but rather locatum and relatum

owl/EASE-SAY.owl
daniel86: change to: *A schematic theory of a functional relation ....*

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
...
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
```
daniel86, line 295: so this is a many-to-many relation?
Or another reason to not use *exactly 1* here?

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory of how an agent might use an instrument to prevent access to the interior of a patient.</rdfs:comment>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/EASE-SAY.owl#CrackingTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#CrackingTheory">
```
daniel86, line 305: "Cracking" is really that basic?

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory of how an agent might use an instrument to prevent access to the interior of a patient.</rdfs:comment>
```
daniel86, line 298: What is the meaning of "might" in this sentence? Maybe replace with "a plausible way of doing it"?

owl/EASE-SAY.owl
```diff
@@ -113,6 +244,83 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory of how an agent might go about inflicting damage to the surface of an object.</rdfs:comment>
```
daniel86, line 319: What is the meaning of "might" in this sentence? Maybe replace with "a plausible way of doing it"?

owl/EASE-SAY.owl
```diff
@@ -123,6 +331,75 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#PhysicalObjectRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:cardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:cardinality>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The ESTSchema constitutes the most basic schema that describes the existance of an entity, thereby having one Role, i.e. what entity is asserted to exist.
```
daniel86, line 350: rephrase first sentence:

*A schematic theory that describes the **existence** of an entity, ...*

daniel86, line 350: I do not get the part "thereby having one Role". Who has one role, and why? That is not clear to me.

owl/EASE-SAY.owl
```diff
@@ -123,6 +331,75 @@
...
+
+In SOMA, this schema will refer to the existence of physical objects. As such, it is not associated to a relation between two objects.</rdfs:comment>
```
daniel86, line 352: change "will refer" to "refers"

owl/EASE-SAY.owl
daniel86, line 350: what does *EST* stand for? it is written nowhere

owl/EASE-SAY.owl
```diff
@@ -123,6 +331,75 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#SchematicTheory"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#AgentRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>Also known as &quot;executing schemas&quot; or &quot;x-schemas&quot;, these were defined by Bergen and Chang in their work &quot;Embodied Construction Grammar in Simulation-Based Language Understanding&quot; as:
+
+&quot;Executing schemas, or x-schemas, are dynamic representations motivated in part by motor and perceptual systems (Bailey 1997; Narayanan 1997), on the assumption that the same underlying representations used for executing and perceiving an action are brought to bear in understanding language about that action. The
```
daniel86, line 369: formatting seems strange here, better put the quote in just one line.

owl/EASE-SAY.owl
```diff
@@ -123,6 +331,75 @@
...
+x-schema formalism is an extension of Petri nets (Murata 1989) that can model sequential, concurrent, and
+asynchronous events&quot;
+
+In SOMA, we do not restrict the formalism of ExecutableSchematicTheories; i.e. they don&apos;t have to be Petri Nets.
```
daniel86, line 373: Change to "SOMA does not restrict ..." ?

owl/EASE-SAY.owl
```diff
@@ -123,6 +331,75 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-SAY.owl#RelatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The most general type of functional spatial relations.
```
daniel86, line 396: A theory is not a relation, it describes one. Comments should be accurate.

owl/EASE-SAY.owl
```diff
@@ -183,23 +462,17 @@ Note that in a semantic sense such clauses always request information, but in a
...
-        <rdfs:label>Linguistic object</rdfs:label>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#LocatumRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:allValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Denotes the object with primary focal prominence in a spatial or spatio-temporal schema. Terminological variants: Figure (Talmy 1983), Trajector (Langacker 1986).</rdfs:comment>
```
daniel86, line 475: The last sentence is not really a sentence. Please revise it.

owl/EASE-SAY.owl
```diff
@@ -211,20 +484,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/EASE-OBJ.owl#Patient"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory for how an agent can mix a fluid or particulate object.</rdfs:comment>
```
daniel86, line 511: A schematic theory **for** or **about**?

owl/EASE-SAY.owl
```diff
@@ -211,20 +484,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/EASE-SAY.owl#PhysicalObjectRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:allValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The PhysicalObjectRole requires its filler simply to exist, unlike other roles that may demand e.g. agentive or instrumental participation in some executable schema or plan (AgentRole and Instrument respectively).
```
daniel86, line 535: Change beginning to:

"A role that requires ..."

No need to repeat the role name in the comment.

owl/EASE-SAY.owl
```diff
@@ -211,20 +484,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+
+The intention of such a simple role is to have a way to represent, in a schematic formalism used to describe some scene, that an object is present. In particular, the schema used to convey this information is the ESTSchemaTheory, which has PhysicialObjectRole as its sole defined role.</rdfs:comment>
```
daniel86, line 537: **Roles do not have intentions.**
You could write the intention of modelling the role, that would be ok to write.

daniel86:
@mpomarlan you haven't addressed all my comments. I would review again after they have been addressed

owl/EASE-SAY.owl
mpomarlan: For Covering, exactly might work. In general though, several entities may fill the patient role.

owl/EASE-SAY.owl
mpomarlan: Several patients could be possible at once.

owl/EASE-SAY.owl
mpomarlan: I for one don't think so, and I'm not the one that put that in. I think the reason why that's there, is an investigation by Vanja on frequent verbs in cooking recipes, cracking being among the top 20.

owl/EASE-SAY.owl
mpomarlan: That seems needlessly convoluted. "might" in this sentence simply conveys ability; this is a standard, if somewhat rarer than "can", English use in how-to instructive text. See for example

"Aged 40? This is how you might afford the 100 year life" (https://www.trustnet.com/News/852555/aged-40-this-is-how-you-might-afford-the-100-year-life/)

"This is how you might organize conference time to maximize the number of students(...)" (https://www.pinterest.com.au/pin/482025966337075719/)

owl/EASE-SAY.owl
mpomarlan: I'm not sure where the confusion comes from. It's pretty clear the schema being defined has a single role.

mpomarlan: I don't know, ask Vanja.

daniel86: " that describes the existance of an entity, thereby having one Role" 
so how does describing the existence implies that there is one role? I do not get it. Please rephrase.

There is also a typo, it is written **existence**

mpomarlan: There is one role because the schema refers to the object existing. There is no need for any other role.

mpomarlan:
Well, a lot of those iterations were, in my opinion, somewhat unnecessary as they refer to extremely nitpicky aspects about how comments are written -- and the nitpicks aren't even always correct. The usage of "might", for example, is fairly standard.

I have so far accomodated the comments regardless, to the best of my ability, but the timing issue is what it is -- people had holiday plans, and at the moment not all the necessary information is available to answer all the comments. There's not many though, and really, if the one thing holding us back is not having "ESTSchema" explained in comments, I think this is a minor issue that can be fixed soon after everyone is back.

owl/EASE-SAY.owl
mpomarlan: The "existance" typo has already been resolved by my latest commit yesterday; as github shows, this convo is on outdated code.

daniel86:
> Well, a lot of those iterations were, in my opinion, somewhat unnecessary as they refer to extremely nitpicky aspects about how comments are written -- and the nitpicks aren't even always correct. The usage of "might", for example, is fairly standard.
> 
> I have so far accomodated the comments regardless, to the best of my ability, but the timing issue is what it is -- people had holiday plans, and at the moment not all the necessary information is available to answer all the comments. There's not many though, and really, if the one thing holding us back is not having "ESTSchema" explained in comments, I think this is a minor issue that can be fixed soon after everyone is back.

glad that you classify my efforts to do multiple iterations of review even though I have a lot of other things todo as "unnessary". Howver, not very kind.
And no, I haven't finished the review as I have stopped again after I had a bunch of comments written down. I haven't looked deeply into axiomatization in the pull request yet.

mpomarlan:
From my side, a lot of these comments look like stalling. Rephrasing comments is something that can be easily fixed if needed, after this is merged and the refactoring happens.

daniel86:
> From my side, a lot of these comments look like stalling. Rephrasing comments is something that can be easily fixed if needed, after this is merged and the refactoring happens.

well all can easily be fixed afterwards, but why?
My procedure is first making the comments nice, so that I understand all the concepts from looking at the comments.
And second look into whether the axioms match the definition given in the comment.
So for me comments are very important, and should be worked on first before considering axiomatization -- because you need to know exactly what you want to formally define

mpomarlan:
I think something good can get excessive.

owl/SOMA-SAY.owl
```diff
@@ -6,9 +6,12 @@
      xmlns:xml="http://www.w3.org/XML/1998/namespace"
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
      xmlns:SOMA="http://www.ease-crc.org/ont/SOMA.owl#"
-     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
+     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+     xmlns:swrl="http://www.w3.org/2003/11/swrl#"
+     xmlns:knowrob="http://knowrob.org/kb/knowrob.owl#">
```
daniel86, line 11: I wonder where this is coming from?

owl/SOMA-SAY.owl
```diff
@@ -80,13 +166,69 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#CausalProcessRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#classifies"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ExecutableSchematicTheory"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A role that denotes physical processes causing a motion such as the force interaction of a tennis racket hitting a ball and setting it into motion.</rdfs:comment>
```
daniel86, line 179: the comment is a bit confusing. For me, "A role that denotes physical processes" implies that the role is used to classify processes. Which would not be consistent as processes are no objects. And the axiomatization states that the role classifies some theory.
Maybe could be clarified?

owl/SOMA-SAY.owl
```diff
@@ -80,13 +166,69 @@
...
+
     <!-- http://www.ease-crc.org/ont/SOMA.owl#ClausalObject -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#ClausalObject">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#LinguisticObject"/>
-        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#LexicalObject"/>
-        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#PhrasalObject"/>
-        <rdfs:comment>A linguistic object that contains information which would be linguistically expressed as a clause.</rdfs:comment>
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/IOLite.owl#Phrase"/>
+        <rdfs:comment>A linguistic object that contains information which would be linguistically expressed as a clause.
```
daniel86, line 229: I would suggest to remove this sentence. The second sentence seems sufficient to me

owl/SOMA-SAY.owl
```diff
@@ -121,6 +341,74 @@
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#PhysicalObjectRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:cardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:cardinality>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A schematic theory that describes the existence of an entity, thereby having one Role, i.e. what entity is asserted to exist.
```
daniel86, line 360: I am still a bit puzzled about this sentence.
So well, an object exists within its lifetime (which is an event).
It surely participates in this event, but it is not clear to me how and why an the participation of the object would be classified.
The last piece states that this role is "what entity is asserted to exist". How is that a role?

Maybe some illustration would help here? Or more details about the lifetime event and particpation

owl/SOMA-SAY.owl
```diff
@@ -121,6 +341,74 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#ExecutableSchematicTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#SchematicTheory"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#AgentRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>Also known as &quot;executing schemas&quot; or &quot;x-schemas&quot;, these were defined by Bergen and Chang in their work &quot;Embodied Construction Grammar in Simulation-Based Language Understanding&quot; as:
```
daniel86, line 377: - I think the comment is a bit long
- I think you should start by stating what it is and not with aka. -- it is also a bit redundant as you repeat both in nextparagraph

owl/SOMA-SAY.owl
```diff
@@ -209,20 +493,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Instrument"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Patient"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory about how an agent can mix a fluid or particulate object.</rdfs:comment>
```
daniel86, line 520: no *might* here? I would vote for removing it in the other comments too

owl/SOMA-SAY.owl
```diff
@@ -209,20 +493,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#PathRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#PathRole">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Role"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The path of a motion is represented by a Source-Path-Goal schema.</rdfs:comment>
```
daniel86, line 529: change to:

*A role that classifies the path of a motion ....*

owl/SOMA-SAY.owl
```diff
@@ -209,20 +493,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#PhysicalObjectRole -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#PhysicalObjectRole">
```
daniel86, line 536: The name of the role seems not intuitive to me.
Could we find something more obvious? like *Existent* or so?

owl/SOMA-SAY.owl
```diff
@@ -209,20 +493,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Destination"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Patient"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>A schematic theory of how an agent might set a patient at a particular goal location.</rdfs:comment>
```
daniel86, line 567: maybe better: *... how an agent might place ...* ?

owl/SOMA-SAY.owl
```diff
@@ -209,20 +493,107 @@ Note that in a semantic sense such clauses always request information, but in a
...
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#LocatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#defines"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#RelatumRole"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A qualitative spatial relation indicating relative proximity, as expressed by the prepositions &apos;near&apos;, &apos;close to&apos;, &apos;next to&apos;, etc.
```
daniel86, line 588: change to:

*An image schematic theory that describes a qualitative ....*

owl/SOMA-SAY.owl
```diff
@@ -308,6 +725,37 @@ Note that such theories are useful both when analyzing an actual linguistic prod
...
         <rdfs:comment>A transition between two states brought about by the Action of some Agent.</rdfs:comment>
     </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#SupportTheory -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#SupportTheory">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ControlTheory"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The reified functional relation holding between two spatial objects x and y, such that x physically supports y in the presence of gravity; x and y need not be in contact. An example of such an expression is &quot;The bowl is on the table&quot;.
```
daniel86, line 735: change to:

An image schematic theory that describes ....

owl/SOMA-SAY.owl
daniel86: The comment for the role is much more clear, in my opinion

owl/SOMA-SAY.owl
daniel86: another thought: isn't it the case that social objects can also stop existing? i.e. when the knowledge is "forgotten"

mpomarlan: I think a colloquial way to understand the idea behind "object permanence" in psychology (and the corresponding ESTSchemaTheories in SOMA) is the good old proverb "reality is that which, when you stop believing in it, doesn't go away".

As you say, Social Objects are obviously very dependent on the observer (though even there it is more related to the consensus of observers -- I can stop believing in money, but the German government will still demand I pay taxes). But I think, ultimately, we don't want to go there with this role and schema pair. We really aim to capture "(physical) object permanence", which is a very basic, but apparently not innate or at least not immediately manifested, cognitive ability.

daniel86:
> I think the request can be merged. But let us first include SOMA-SAY in CI -- currently it is not tested in CI whether SOMA-SAY has inconsistencies.
> 
> Could you maybe include enabling it for CI in this pull request?
> 
> you can activate it here: https://github.com/ease-crc/soma/blob/master/prolog/uglify.pl#L63

I included it, let's hope hermit is happy :)

mpomarlan:
Yay, thanks!

Yes, let's see about HermiT. Unfortunately, ontologydesignpatterns (the website) seems down at the moment, so HermiT cannot find DUL. We may have to refresh the HermiT check some other time when odp gets back.

daniel86:
> Yay, thanks!
> 
> Yes, let's see about HermiT. Unfortunately, ontologydesignpatterns (the website) seems down at the moment, so HermiT cannot find DUL. We may have to refresh the HermiT check some other time when odp gets back.

oh, good catch! ok let's wait

PR #92, opened 2020-07-17
SOMA refactoring

daniel86:
As discussed last week, this pull request addresses unifying the IRI namespace over all OWL files.
This includes:
- rename OWL files from "EASE*" to "SOMA*"
- xml base for every OWL file is "http://www.ease-crc.org/ont/SOMA.owl"
- all SOMA entities are defined in the namespace "http://www.ease-crc.org/ont/SOMA.owl#"
- additionally, deleted all "package://" imports and replaced them with proper http imports


daniel86:
probably not really possible to review, and it will cause conflicts with the other pending pull request. But we have discussed to do the refactoring first, before anything else. And it might take some more time to get the SOMA-SAY pull request through

mpomarlan:
If I remember, no, we said we merge the SAY pull request and then this one, precisely to avoid breaking git's ability to merge.

daniel86:
> If I remember, no, we said we merge the SAY pull request and then this one, precisely to avoid breaking git's ability to merge.

nope, definitely not. We said we merge the design request, and then we do the refactoring.

mpomarlan:
That's not how I remember it. The SOMA-SAY was opened 13 days ago when the Design PR was merged, because we all knew it was in the pipeline and it would make sense to get it in before the big refactoring.

daniel86:
> That's not how I remember it. The SOMA-SAY was opened 13 days ago when the Design PR was merged, because we all knew it was in the pipeline and it would make sense to get it in before the big refactoring.

well then you remember it wrong.
We discussed in the meeting before I merged the then opened design request to only merge this one, then do the refactoring. I am 100% sure.

mpomarlan:
I don't know why we would have agreed to such a thing, when we knew SOMA-SAY had a PR in the pipeline and that merging it will be more difficult after the refactoring.

daniel86:
> I don't know why we would have agreed to such a thing, when we knew SOMA-SAY had a PR in the pipeline and that merging it will be more difficult after the refactoring.

yeah well we cannot change that. So let's wait until everyone is back.
I'll be not available for two weeks now.

mpomarlan:
The idea was to get the SAY PR in while Vanja was still here.

daniel86:
> The idea was to get the SAY PR in while Vanja was still here.

Well if you say so. I can only repeat that I am 100% sure that you are wrong

owl/SOMA-ELAN.owl
```diff
@@ -0,0 +1,358 @@
...
+        <SOMA:ELANUsageGuideline>walk (to a location, or in a direction)</EASE:ELANUsageGuideline>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://www.ease-crc.org/ont/SOMA.owl#BodyMovement">
+        <SOMA:ELANName>move-body-general</EASE:ELANName>
+        <SOMA:ELANUsageGuideline>The Agent moves its body in some way. Use more specific labels where appropriate.</EASE:ELANUsageGuideline>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://www.ease-crc.org/ont/SOMA.owl#HeadMovement">
+        <SOMA:ELANName>move-head</EASE:ELANName>
```
mpomarlan, line 64: Closing tag and opening tags don't match.

Issue #94, opened 2020-07-20
Representing Slicing Actions


joserochh:
We were discussing on what could be a good fit to represent slicing actions:

EASE-ACT.owl#Actuating          focus more in the movement of the object
EASE-ACT.owl#Manipulating    focus more in the movement of the hand
Or rather if there should be a special class for it?

Slicing could be thought, for example, on how you manipulate a knife to achieve a cutting. Or how the knife is acted to achieve the cutting, specially because a blade can cut without being manipulated. Maybe in the specific example of slicing something with your hand both Actuating and Manipulating are involved.

mpomarlan:
Hi!

This is an interesting case indeed. One can argue that the important thing about such an action isn't that it is performed by hand (as a manipulation would) or that it involves moving an object outside of the agent (as actuation would), but its effect on some other object. Specifically, slicing is that which, at least partially, transforms something into slices; it may involve a tool.

We have some other kinds of actions (or tasks) defined in our ontology, and in particular ModifyingPhysicalObject seems like a good superclass for Slicing.

At the moment, some of us from the onto dev team are on holiday for a few days, so expect this to not be done tomorrow, but I'll open a pull request to fix this in the beginning of next week.

PR #96, opened 2020-07-27
Added cutting, slicing, dicing.

mpomarlan:
Solves issue #94

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -179,28 +179,28 @@
     
 
 
-    <!-- http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting -->
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#Cutting -->
 
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting">
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Cutting">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ModifyingPhysicalObject"/>
-        <rdfs:comment>The goal of this task is to separate some target object into several pieces by means of cutting into its constituent material. Unlike a disassembly, a cut is usually not easily reversible.</rdfs:comment>
+        <rdfs:comment>The goal of this task is to separate one or more pieces from some target object by means of cutting into its constituent material. Unlike a disassembly, a cut is usually not easily reversible.</rdfs:comment>
     </owl:Class>
     
 
 
-    <!-- http://www.ease-crc.org/ont/SOMA-ACT.owl#Dicing -->
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#Dicing -->
 
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-ACT.owl#Dicing">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting"/>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Dicing">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Cutting"/>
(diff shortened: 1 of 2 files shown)


owl/SOMA-ACT.owl
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0"?>
-<rdf:RDF xmlns="http://www.ease-crc.org/ont/SOMA.owl#"
-     xml:base="http://www.ease-crc.org/ont/SOMA.owl"
-     xmlns:SOMA="http://www.ease-crc.org/ont/SOMA.owl#"
-     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+<rdf:RDF xmlns="http://www.ease-crc.org/ont/SOMA-ACT.owl#"
+     xml:base="http://www.ease-crc.org/ont/SOMA-ACT.owl"
```
daniel86, line 3: Protege has changed back the xml:base :(
How can we avoid that?
It must be that it overwrites it with the ontology rdf:about?

daniel86: I guess most pragmatic and easy would be to change all ontology rdf:about to SOMA.owl

daniel86: this would change all the rdf:about's: https://github.com/ease-crc/soma/pull/97

daniel86: does not seem like an option because Protege would not load two modules next to each other anymore :(

daniel86: ok, change of plan. part of the refactoring must happen in the uglify step

owl/SOMA-ACT.owl
```diff
@@ -179,6 +179,33 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ModifyingPhysicalObject"/>
+        <rdfs:comment>The goal of this task is to separate some target object into several pieces by means of cutting into its constituent material. Unlike a disassembly, a cut is usually not easily reversible.</rdfs:comment>
```
daniel86, line 186: several -> one or more

owl/SOMA-ACT.owl
```diff
@@ -179,6 +179,33 @@
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA-ACT.owl#Dicing -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-ACT.owl#Dicing">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA-ACT.owl#Cutting"/>
+        <rdfs:comment>A particular kind of cutting where the goal is to produce small pieces out of some object or material. Unlike slices, the pieces to be obtained do not have one or two dimensions being more prominent than others. &quot;Dice&quot;, the pieces dicing results in, are approximately cubic.</rdfs:comment>
```
daniel86, line 195: at a later point we could axiomatize the cubic shape

mpomarlan:
I changed the concept IRIs for Cutting, Slicing, Dicing to be SOMA.owl rather than SOMA-ACT.owl

As you noticed, changing the xml:base etc tags wrecks Protege's ability to actually import dependencies, so for now I kept the xml:base for SOMA-{ACT,ELAN} to be also SOMA-{ACT,ELAN}.

I suggest we revert the xml:base and xmlns in every other file as well.

An observation -- though not the most ergonomic thing ever, it is possible to customize Protege's entity creation preferences and specify a particular IRI for new entities. It's a possible workaround.

daniel86:
> I changed the concept IRIs for Cutting, Slicing, Dicing to be SOMA.owl rather than SOMA-ACT.owl
> 
> As you noticed, changing the xml:base etc tags wrecks Protege's ability to actually import dependencies, so for now I kept the xml:base for SOMA-{ACT,ELAN} to be also SOMA-{ACT,ELAN}.
> 
> I suggest we revert the xml:base and xmlns in every other file as well.
> 
> An observation -- though not the most ergonomic thing ever, it is possible to customize Protege's entity creation preferences and specify a particular IRI for new entities. It's a possible workaround.

yes I realized that as well. sorry for the confusion.
It does not matter anymore as long as the IRI starts with "http://www.ease-crc.org/ont/SOMA".
This is because the uglify step does change all such IRIs to the one SOMA IRI automatically.
So, yes, we can set the xmlns back in all files.

PR #98, opened 2020-07-27
uglify all SOMA iri's

daniel86:
convert any 'http://www.ease-crc.org/SOMA[.*].owl#[.*]' IRI to 'http://www.ease-crc.org/SOMA.owl#[.*]' in the uglify step


daniel86:
I also included EASE-SAY in the uglify step -- seems there are some inconsistencies

PR #100, opened 2020-08-08
deploy tagged releases

daniel86:
extends CI to auto-deploy on release


daniel86:
I took the freedom to merge without review as this only affects CI

PR #103, opened 2020-08-11
Checking possible inconsistencies in SOMA-STATE

mpomarlan:


mpomarlan:
I think I found the problem. A concept called Containment was used both as a Disposition (ie. Quality) and as StateType (ie. Object), and was therefore equivalent to Nothing. This propagated to some other classes via existential concept constructions and subsumption axioms.

The proposed fix in this pull request: Containment names a Disposition. The state type is now ContainmentState. Analogously, we have a SupportState instead of the previous Support. Each of these state types also links to the relevant disposition, via the isStateOf property. 

This property links a state type to a role. We now have that ContainmentState must link to a Container role that classifies only objects with the Containment disposition. Similarly, SupportState must link to a role that classifies only objects with the Deposition disposition.

mpomarlan:
Note on why the isStateOf axiom simply existential, as opposed to a pair of existential+closure axiom (isStateOf only etc): the state, eg. ContainmentState, can affect several objects, such as container and containees.

owl/SOMA-STATE.owl
```diff
@@ -104,39 +104,32 @@
...
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Restrictor"/>
-    
-
+    <!-- http://www.ease-crc.org/ont/SOMA-STATE.owl#ContainmentState -->
 
-    <!-- http://www.ease-crc.org/ont/SOMA.owl#Supporter -->
-
-    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Supporter"/>
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-STATE.owl#ContainmentState">
```
daniel86, line 109: comment is missing

Issue #105, opened 2020-08-15
Action Taxonomy


daniel86:
- Change subclasses of Action to only PhysicalAction, CommunicationAction and MentalAction

PR #106, opened 2020-08-15
action cleanup

daniel86:
this pull request fixes issue  #105


daniel86:
> One comment I'd like to discuss before I approve this is the new concept PhysicalAction. The comment suggests it is meant to replace the older ManipulationAction concept, but I don't think this is appropriate because not all PhysicalActions are manipulations. Manipulation actions are physical actions involving certain parts of an agent as participants, but there can be physical actions without the involvement of those parts.

yes you are right, I will update the comment

PR #107, opened 2020-08-16
First version of SOMA webpage

AbhijitVyas:
you can find it online as of now on https://abhijitvyas.github.io/soma/


daniel86:
fancy ;)

One thing missing: allow to load different SOMA versions into the ontology browser.
I think somewhere should be a selection box, where the user can pick a SOMA version.
Default would be the latest version.

Also we will need to fill it with something else then blindtext to merge it into master.
But we can stick to this layout, and now fill it with content.
We can also do this before picking a concept browser and stick to webvowl for now.

daniel86 (changes requested):
looks good, now we need to replace the blindtext.
To merge it, it must be a first complete version of the page.

daniel86 (changes requested):
please remove auto-generated files in docs/NEEM-specification
such as main.log -- it is one of the files created by latex and should not be added to git

daniel86 (approved):

PR #123, opened 2020-09-06
Physical object branch cleanup

AbhijitVyas:


owl/SOMA-OBJ.owl
```diff
@@ -1548,7 +1535,7 @@ The paradigmatic example is the Agent performing an Action -- the Agent is the e
     <!-- http://www.ease-crc.org/ont/SOMA.owl#DesignedComponent -->
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#DesignedComponent">
-        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#FunctionalPart"/>
```
daniel86: you mix up stuff here :/

owl/SOMA-OBJ.owl
```diff
@@ -975,6 +975,9 @@
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#ArtificialAgent">
         <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalAgent"/>
+        <rdfs:comment>A physical objects with artificial characteristics, which can perform actions to achieve desired goals, typically has sensors and/or actuators.
```
daniel86, line 978: you mix up two pull requests here.
better use two *separate* branches in case you do two separate things

daniel86 (changes requested):

owl/SOMA-OBJ.owl
```diff
@@ -2973,17 +2257,6 @@ The intention behind such a simple role is to have a way to represent, in a sche
                 </owl:allValuesFrom>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
```
daniel86, line 2976: here is a deleted axiom.
Probably it would be good to keep it and change *dirty object* to *physical object*

daniel86 (changes requested):
the cut-off point seems reasonable.
There is one (accidentally) deleted axiom which should be rewritten instead, otherwise I would accept the reuest

daniel86 (changes requested):
the axiom is still deleted, please re-add the axiom

daniel86 (approved):

PR #127, opened 2020-09-09
Shape region: now scalable and localizable.

mpomarlan:
This answers some stuff we discovered while writing the NEEM doc: shapes should be scalable (this includes flippable) and localizable. Further, there is now a specialized property linking physical objects to their shapes.

Also, uses hasLength as a property for boxes, rather than hasDepth.


mpomarlan:
During comparison of this to master, I noticed many trivial differences: either the master file or my version includes empty descriptions of DUL concepts, tagged as either owl:Class or rdf:Description.

Clearly, Protege is adding these for some reason, but the different tags suggest different versions. Something to think about.

daniel86:
> During comparison of this to master, I noticed many trivial differences: either the master file or my version includes empty descriptions of DUL concepts, tagged as either owl:Class or rdf:Description.
> 
> Clearly, Protege is adding these for some reason, but the different tags suggest different versions. Something to think about.

it seems you have used an older version of Protege, which is noticeable by the OWL API version inserted in the OWL file.
Maybe this has caused problems?

Some time ago, Sascha has merged in the latest protege version into our fork: https://github.com/daniel86/ros-protege/
Maybe it would save some trouble to use this one.

daniel86:
@mpomarlan unfortunately merging in Abhijit changeset has created a conflict here

daniel86 (changes requested):
could you fix the conflict?
Sorry for the review delay!

mpomarlan:
I think the conflict is gone now. Have a look.

owl/SOMA-OBJ.owl
```diff
@@ -24,6 +24,31 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasShape -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasShape">
```
daniel86, line 29: There is a nameclash: *hasShape* is already defined as a sub-property of *hasQuality*.

Would it be ok to name this property *hasShapeRegion*? Then the domain could be `(Shape or PhysicalObject)`.

owl/SOMA-OBJ.owl
```diff
@@ -787,6 +812,36 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasReferencePose -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasReferencePose">
+        <rdfs:subPropertyOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#hasShapeParameter"/>
+        <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ShapeRegion"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
daniel86, line 820: Can we refer to a space region instead?
I think there is e.g. `6DPose` which is a region.
Maybe this can be modeled as part or constituent?

owl/SOMA-OBJ.owl
```diff
@@ -787,6 +812,36 @@
...
+        <rdfs:comment>Gives the default pose of a shape in its own coordinate system.
+
+It is sometimes the case that a shape described via a mesh or a primitive needs to be translated and/or rotated in order to correspond to the object it is a shape of. This can happen, e.g., when a robot description file places the center point of a link at a different location to the centroid of the mesh giving that link&apos;s shape.</rdfs:comment>
```
daniel86, line 823: in case of sticking to string values, I think it would be good to write what the format of poses is, e.g. `[frame,position,quaternion]`

owl/SOMA-OBJ.owl
```diff
@@ -787,6 +812,36 @@
...
+    </owl:DatatypeProperty>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasShapeScale -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasShapeScale">
+        <rdfs:subPropertyOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#hasShapeParameter"/>
+        <rdfs:domain rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ShapeRegion"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
daniel86, line 834: Could we use three individual properties with xsd:double range instead (or in addition)?

We could very well also expand the representation of pose data in my opinion. The reason being that the data is not actually stored in the triple store (anymore), but rather in a separate DB collection for TF data, or in URDF files. So the model defined here would mainly be used by KnowRob to access this externally stored data.

owl/SOMA-OBJ.owl
```diff
@@ -2431,6 +2493,20 @@ Resources are often consumed by their participation in an event, but this need n
...
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasReferencePose"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasShapeScale"/>
```
daniel86, line 2505: In KnowRob, the scale is only a property of mesh shapes.
Do we need scale for anything other then the mesh shapes?

daniel86 (changes requested):

owl/SOMA-OBJ.owl
mpomarlan: True. The same format should be used as for 6DoF pose region.

owl/SOMA-OBJ.owl
mpomarlan: In principle, it is possible to apply scaling to any geometric primitive. Of course, this can result in the shape being no longer the initial primitive (e.g. a sphere becomes an ellipsoid). As far as I can tell, for example RViz uses scale.x, scale.y, scale.z parameters in a marker message even for some of the basic shapes such as the cylinder.

OTOH, we don't need to do this. Cylinders can be specified by radius, height; spheres are not ellipsoids. So to keep things simple we can restrict scaling to meshes.

mpomarlan:
I had a look through the comments, they can be implemented more or less as-is. I would like though to ask that we keep scaling as an xsd:string, in the same way that vectors are currently represented. If there will be a refactoring of how 3DVector or 6DPose are to be represented, let's think about it a bit more.

owl/SOMA-OBJ.owl
daniel86: Yes you are right.

It started with early KnowRob representing 4x4 matrices in the triple store with 16 different properties, one for each slot in the matrix.
That I still think is very excessive and not very meaningful.
I first wanted to avoid these unneeded triples that would not appear in any axiom anyways.
Then I tried to introduce these vector datatypes which are still in SOMA.owl, I think.
But that was never really clean, I would say.

The situation has changed a bit since KnowRob stopped to actually store this information in the triple DB, but rather uses external sources to draw e.g. pose data from TF or shape data from URDF.
So the argument of doing it like that for performance reasons is not entirely valid anymore.

I agree that we should try to be as consistent as possible in how we handle data.
for example, box attributes have at the moment their own property each (i.e. for width, height etc.).
For color and scale we use strings that encode a vector instead, and for poses we have a string attribute that encodes the whole pose.

Surely one disadvantage of strings is that the model does not enforce format, and that applications would need to have their own parser for these strings (in case of not using KnowRob to read the data).

daniel86:
> I had a look through the comments, they can be implemented more or less as-is. I would like though to ask that we keep scaling as an xsd:string, in the same way that vectors are currently represented. If there will be a refactoring of how 3DVector or 6DPose are to be represented, let's think about it a bit more.

ok.
Have you had a look at the vector datatypes? Using them has disadvantage that we leave xsd, but we could make more explicit what data is in there. And I could have a parser in knowrob that acts upon the datatype (which is not possible when using xsd:string).
Also, then the shape properties should be merged into one vector?

owl/SOMA-OBJ.owl
daniel86: looking at SOMA.owl `6DPose` uses three data properties `hasOrientationVector`, `hasTranslationVector` and `hasReferenceFrame`

owl/SOMA-OBJ.owl
mpomarlan: Sure!

mpomarlan:
Vector datatypes? Sounds cool, but can you give me an example?

As for the shape properties, I think those should not be merged into a vector. It so happens that in the case of the box you can get away with it, but there's not a mathematical sense in which a space of radius-height pairs (the parameters of a cylinder) would be a vector space.

Scale otoh can be a vector.

daniel86:
> Vector datatypes? Sounds cool, but can you give me an example?
> 
> As for the shape properties, I think those should not be merged into a vector. It so happens that in the case of the box you can get away with it, but there's not a mathematical sense in which a space of radius-height pairs (the parameters of a cylinder) would be a vector space.
> 
> Scale otoh can be a vector.

have a look in protege, you can seect the vector datatypes instead of xsd types. e.g. `array_double` is one.
Probably their definition would need a revision, but well, they are there

daniel86:
@mpomarlan  should I review again, or wait until you considered the array types?

mpomarlan:
Give me a minute, I'll update the vector type.

mpomarlan:
Ok, replaced the type to the array_double.

daniel86 (approved):

PR #129, opened 2020-09-13
Document soma - Glossary

kaviyachandran:
Hello @daniel86 ,

I modified the make file to use pdflatex. Pythontex does not work with latexmk (it might with a .latexmkrc). Is it okay to use pdflatex or should I use latexmk. I want a review before merge so creating a PR. But still there are some problems to fix. owl ready does not read the comments of some entities.


scripts/owl_reader.py
```diff
@@ -0,0 +1,104 @@
+import owlready2 as owl
+import time
+import sys
+from collections import OrderedDict
+import re
+
+class OwlReader:
```
daniel86, line 7: could you add a small comment to this file about what the purpose of the script is?

daniel86 (changes requested):

scripts/owl_reader.py
kaviyachandran: Sure

daniel86 (approved):

PR #134, opened 2020-09-19
added URDF-like model for joints

daniel86:
This pull requests adds a few new concepts and properties to model kinematics.
- joints are physical bodies with child and parent link
- there are different types of joints
- each joint has a joint state attribute
- some joints may have physical limits
- components of objects may be root or leaf of the kinematic structure of the object

### DIFF
--- a/owl/SOMA-OBJ.owl
+++ b/owl/SOMA-OBJ.owl
@@ -1256,6 +1256,13 @@
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#FixedJoint">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Joint"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasJointState"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">0</owl:qualifiedCardinality>
+                <owl:onClass rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Entity"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
         <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#MovableJoint"/>
         <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A joint that cannot move, designed to fixiate links.</rdfs:comment>
         <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Fixed joint</rdfs:label>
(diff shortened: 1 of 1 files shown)


owl/SOMA-OBJ.owl
```diff
@@ -971,6 +1242,173 @@
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasJointVelocity"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#double"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The state of a joint in terms of position, velocity of the joint and effort applied to it.</rdfs:comment>
```
mpomarlan, line 1352: Shouldn't there be some axiom about the effort property?

owl/SOMA-OBJ.owl
```diff
@@ -971,6 +1242,173 @@
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasParentLink"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onClass rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">An object that is used to articulate links in a kinematic structure.</rdfs:comment>
```
mpomarlan, line 1319: Does a fixed joint have a joint state? Because of ow semantics, we can say yes and not supply one, but this leaves it possible in our model to have changing joint states to a fixed joint.

owl/SOMA-OBJ.owl
```diff
@@ -971,6 +1242,173 @@
...
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#RevoluteJoint">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#HingeJoint"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasJointLimit"/>
+                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
+                <owl:onClass rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#JointLimit"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A hinge joint that rotates along an axis and has a limited range specified by the upper and lower limits.</rdfs:comment>
```
mpomarlan, line 1406: The terminology I'm familiar with has hinge and revolute as two names for the same thing, rather than one be the subclass of the other.

mpomarlan (changes requested):

owl/SOMA-OBJ.owl
daniel86: To be honest, I am not sure.
Here is the JointState message: http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html
Probably I left effort out because in there they explicitly write about joints without associated effort, but I am not sure what this means -- is this just that it cannot be measured?

mpomarlan: They say every field of the joint state (position, velocity, effort) is optional-- you can supply whatever data you have and ignore what you don't.

I think something like a "max 1" axiom would be a good compromise.

mpomarlan: The text as written suggests yes -- supply whatever data you have etc. But we don't have to do that. We can keep exactly 1 position, exactly 1 velocity, max 1 effort to harmonize with the comment you wrote in the owl file as well.

daniel86: done

owl/SOMA-OBJ.owl
daniel86: I have asserted the following now:

- FixedJoint 'has joint state' exactly 0 Entity
- MovableJoint 'has joint state' exactly 1 'Joint state'

mpomarlan:
FixedJoint is now equivalent to Nothing. This might be because its parent says there should be some jointstate attached to a joint, whereas FixedJoint says there should be none.

daniel86:
> FixedJoint is now equivalent to Nothing. This might be because its parent says there should be some jointstate attached to a joint, whereas FixedJoint says there should be none.

sorry, I will wait now until I see results before I request review again

daniel86:
@mpomarlan ok it should be ready now

mpomarlan (approved):

PR #136, opened 2020-09-19
Mass and NetForce attributes

daniel86:
- added missing inverse of `hasMassAttribute`
- added `NetForce` concept, `hasNetForce` and `hasForceValue` properties

### DIFF
--- a/owl/SOMA-OBJ.owl
+++ b/owl/SOMA-OBJ.owl
@@ -24,6 +24,19 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasNetForce -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasNetForce">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRegion"/>
+        <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#isNetForceOf"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <rdfs:range rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#NetForce"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The accumulated force acting upon an object.</rdfs:comment>
+        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">has net force</rdfs:label>
+    </owl:ObjectProperty>
+    
+
+
     <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#isMassAttributeOf -->
 
     <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#isMassAttributeOf">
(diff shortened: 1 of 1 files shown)


owl/SOMA-OBJ.owl
```diff
@@ -24,6 +24,44 @@
...
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasNetForce -->
+
+    <owl:ObjectProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasNetForce">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRegion"/>
+        <owl:inverseOf rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#isNetForceOf"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalObject"/>
+        <rdfs:range rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#NetForce"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The accumulated force acting upon an object.</rdfs:comment>
```
mpomarlan, line 34: Perhaps something like "A relation between a physical object and the total force acting on it."

owl/SOMA-OBJ.owl
```diff
@@ -790,6 +828,18 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasForceValue -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasForceValue">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRegionDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Region"/>
```
mpomarlan, line 835: Should this property have any region as domain?

owl/SOMA-OBJ.owl
```diff
@@ -971,6 +1021,23 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#NetForce -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#NetForce">
+        <rdfs:subClassOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#PhysicalAttribute"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasForceValue"/>
```
mpomarlan, line 1030: hasForceValue has xsd:double as range. This covers the force magnitude, but what about its direction?

mpomarlan (changes requested):
How will the different forces at different times be represented? Temporalized triples?

owl/SOMA-OBJ.owl
daniel86: Well one could compute from the position change, I assume.

owl/SOMA-OBJ.owl
daniel86: I added it intentionally, yes. Because I didn't want to exclude things that are not modeled. So this net force attribute is in my opinion not the only region that could have a force value. That's why I left it with the general domain axiom

daniel86:
> How will the different forces at different times be represented? Temporalized triples?

yes

owl/SOMA-OBJ.owl
daniel86: ok!

owl/SOMA-OBJ.owl
mpomarlan: Would Color have a force, or Shape? I don't think it will make sense, except perhaps as metaphor but we don't care about those, to have force values for most regions. Would it then make sense to have a ForceRegion class or similar, with NetForce as a subclass?

daniel86: Is every force region an attribute then?

owl/SOMA-OBJ.owl
mpomarlan: Yeah. Net forces in simulators are represented as vectors even, and this representation takes care of both magnitude and direction obviously.

mpomarlan:
Speaking of NetForce being (or also using) a Vector representation, where is this net force applied?

The usual simulator convention is that the net force is applied at the center of mass. This is so often used that I don't think we need to deviate from it, but let's put it in a comment.

owl/SOMA-OBJ.owl
mpomarlan: What's an attribute?

daniel86: `PhysicalAttribute` in DUL is a "Physical value of a physical object, e.g. density, color, etc.".
I have defined NetForce as subclass of PhysicalAttribute

daniel86: Then I would basically define `ForceAttribute` and `NetForce` as a sub-attribute. Then I could add an axiom about hasForceValue only giving a value to force attributes

owl/SOMA-OBJ.owl
mpomarlan: Would then using one of the vector types for the net force value help?

owl/SOMA-OBJ.owl
mpomarlan: That seems like a plan, let's have it so.

mpomarlan (changes requested):

owl/SOMA-OBJ.owl
daniel86: ok, done

mpomarlan (approved):

PR #145, opened 2020-09-24
Added a data property to link DOIs to an InformationObject.

mpomarlan:
It turns out, very little needs to change to support linking info objects to entities.

DUL provides an 'is about' property, with inverse 'is reference of'. The 'is about' property has an InformationObject as domain, and ranges over any entity. This can connect the information object to any individual we have more data about.

DUL also provides an 'expresses' property, with inverse 'is expressed by'. The 'expresses' property has InformationObject as domain, and ranges over social objects such as Concepts or Descriptions. This property therefore can be used to link an InformationObject to what it "means".

I added a comment about InformationObjects to describe this intended pattern of usage, such that it is visible when viewing InformationObject in Protege. This pattern of usage is encoded in the domain/range axioms of the 'expresses' and 'is about' properties.

I also added a new data property to connect the InformationObject to a DOI. Note that in DUL one would more directly connect an InformationRealization to a DOI and an InformationObject, but in the interest of saving a triple it is nice, I think, to have a way of connecting realization DOIs to the InformationObject to which those realizations belong.

### DIFF
--- a/owl/SOMA.owl
+++ b/owl/SOMA.owl
@@ -880,6 +880,17 @@ Sub-properties are used to distinct between different cases of event endpoint re
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationObject"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#anyURI"/>
+        <rdfs:comment>A property linking an InformationObject to a persistent identifier such as a DOI, which can then be used to obtain an address at which a realization (digital file) of the information object can be retrieved.</rdfs:comment>
+    </owl:DatatypeProperty>
+    
+
+
     <!-- http://www.ease-crc.org/ont/SOMA.owl#isReificationOf -->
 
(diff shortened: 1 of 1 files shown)


mpomarlan:
PS: branch can be deleted post-merge.

daniel86 (approved):

daniel86:
@mpomarlan should we also have sub-classes for e.g. VideoFile, MeshFile, URDFFile ?

mpomarlan:
We could have classes of this sort, yes, I was wondering though whether it makes sense to have some more semantics attached. This is where the Description would come in.

But actually that might take a bit of ontological development. So ok, I'll open another pull request in a moment, where I also add the classes you suggest. We can later restrict them to instances of InformationObject that only express certain kinds of Descriptions.

mpomarlan:
With the onto as it is now, linking the URDF might look like this:

PR2_abc isA Robot //let's say
PR2URDF_xyz isA InformationObject
PR2URDF_xyz isAbout PR2_abc
PR2URDF_xyz hasRealizationWithIdentifier "probably/just/the/url/for/now"
PR2URDF_xyz expresses Description_123 //insert haxx here to use Description_123 as a flag for kino-dynamic structure

daniel86:
> We could have classes of this sort, yes, I was wondering though whether it makes sense to have some more semantics attached. This is where the Description would come in.
> 
> But actually that might take a bit of ontological development. So ok, I'll open another pull request in a moment, where I also add the classes you suggest. We can later restrict them to instances of InformationObject that only express certain kinds of Descriptions.

great, thanks

daniel86:
> With the onto as it is now, linking the URDF might look like this:
> 
> PR2_abc isA Robot //let's say
> PR2URDF_xyz isA InformationObject
> PR2URDF_xyz isAbout PR2_abc
> PR2URDF_xyz hasRealizationWithIdentifier "probably/just/the/url/for/now"
> PR2URDF_xyz expresses Description_123 //insert haxx here to use Description_123 as a flag for kino-dynamic structure

I think would be good to have a category attached for handling in knowrob

PR #146, opened 2020-09-24
Added some subclasses of InfoObject

mpomarlan:
Added three classes: VideoData, MeshShapeData, and KinoDynamicData. The latter two are restricted to only be about physical objects. I was not sure whether to restrict VideoData to events. One could say that a video is about an object too, I guess, though I would not be opposed to have VideoData for events only.

The classes are not declared disjoint, but if you want they could be.

### DIFF
--- a/owl/SOMA.owl
+++ b/owl/SOMA.owl
@@ -809,6 +809,7 @@ Sub-properties are used to distinct between different cases of event endpoint re
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationRealization"/>
         <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
+        <rdfs:comment>A property linking an InformationRealization to a string specifying a format name, e.g. URDF or STL.</rdfs:comment>
     </owl:DatatypeProperty>
     
 
--- a/owl/SOMA.owl
+++ b/owl/SOMA.owl
@@ -880,6 +880,17 @@ Sub-properties are used to distinct between different cases of event endpoint re
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationObject"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#anyURI"/>
+        <rdfs:comment>A property linking an InformationObject to a persistent identifier such as a DOI, which can then be used to obtain an address at which a realization (digital file) of the information object can be retrieved.</rdfs:comment>
+    </owl:DatatypeProperty>
+    
+
+
     <!-- http://www.ease-crc.org/ont/SOMA.owl#isReificationOf -->
 
     <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA.owl#isReificationOf">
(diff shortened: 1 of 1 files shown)


owl/SOMA.owl
```diff
@@ -880,6 +880,17 @@ Sub-properties are used to distinct between different cases of event endpoint re
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA.owl#hasRealizationWithIdentifier">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationObject"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#anyURI"/>
```
daniel86, line 888: is every identifier an URI? I think string might be better

daniel86 (changes requested):
My only concern is with range anyURI.
Because for now, I was thinking to just add the filenames there (no path just the name)

A second aspect is the *format*. I think it would be good to have a property for this as different formats might need different handlers

mpomarlan:
Changing to string is no problem.

The specification of a format won't be trivial. Here's a few options I can think of:

1) use subproperties of hasRealizationWithIdentifier. E.g. hasRealizationAsURDFWithIdentifier
2) explicitly represent the InformationRealization individual. Then we could have an InformationObject which expresses something, an InformationRealization which realizes it, and it is the InformationRealization that carries the identifier and format data properties.
3) explicitly represent the InformationRealization, but not the InformationObject. The InformationRealization can then, as above, carry the identifier and format. It can also 'concretely express' some social object e.g. description to say what it means. The 'is about' property would need replacement though with one that connects the InformationRealization to the entity it talks about.

Which one would be better?

daniel86:
> Changing to string is no problem.
> 
> The specification of a format won't be trivial. Here's a few options I can think of:
> 
>     1. use subproperties of hasRealizationWithIdentifier. E.g. hasRealizationAsURDFWithIdentifier
> 
>     2. explicitly represent the InformationRealization individual. Then we could have an InformationObject which expresses something, an InformationRealization which realizes it, and it is the InformationRealization that carries the identifier and format data properties.
> 
>     3. explicitly represent the InformationRealization, but not the InformationObject. The InformationRealization can then, as above, carry the identifier and format. It can also 'concretely express' some social object e.g. description to say what it means. The 'is about' property would need replacement though with one that connects the InformationRealization to the entity it talks about.
> 
> 
> Which one would be better?

I see your point. I am not in favor of 1. I would like that the format itself has an identifier which is a string, e.g., urdf.
Then we would not need to create addition properties/classes for each format.
So I guess between 2. and 3. the question is what do we gain by explication IO? I assume this is a categorization of different IRs, which seems useful to me.

mpomarlan:
So the decision is between 2 and 3 then. Ok.

The problem is that the format is a property of the realization, whereas the referent (the entity the information is about) is considered a property of the information object.

daniel86:
> So the decision is between 2 and 3 then. Ok.
> 
> The problem is that the format is a property of the realization, whereas the referent (the entity the information is about) is considered a property of the information object.

that would be an argument for 2.
well property chain is another option, but well they often cause some trouble for reasoner :)

mpomarlan:
Ok, then I'll do a few adjustments.

daniel86:
Would this be a valid knowledge base:

```
triple(pr2:'PR2_0_Kinematics', rdf:type, soma:'KinoDynamicData'),
triple(pr2:'PR2_0_Kinematics', dul:isAbout, pr2:'PR2_0'),

triple(pr2:'PR2_0_URDF', rdf:type, dul:'InformationRealization'),
triple(pr2:'PR2_0_URDF', dul:realizes, pr2:'PR2_0_Kinematics'),
triple(pr2:'PR2_0_URDF', soma:hasRealizationWithIdentifier, 'pr2'),
triple(pr2:'PR2_0_URDF', soma:hasRealizationFormat, 'URDF')
```

mpomarlan:
On your example knowledge base: yeah, that would be it. Properties are a bit different:

hasPersistentIdentifier
hasDataFormat

but it's the same structure otherwise.

owl/SOMA.owl
```diff
@@ -803,6 +803,16 @@ Sub-properties are used to distinct between different cases of event endpoint re
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#hasDataFormat -->
+
+    <owl:DatatypeProperty rdf:about="http://www.ease-crc.org/ont/SOMA.owl#hasDataFormat">
+        <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasDataValue"/>
+        <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#InformationRealization"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
```
daniel86, line 811: could you add a short comment?

daniel86 (changes requested):
only one comment is missing

daniel86 (approved):

Issue #158, opened 2020-11-04
Mapping CRAM Actions to SOMA Tasks


seba90:
Hi guys,

I created a table with the current CRAM actions which the robot is executes while setting up a breakfast table.
It would be great if we can find official SOMA label for the corresponding CRAM actions which I can later use for the NEEM recording.
The table is structured as follows:

- CRAM Action - The Name of the CRAM Action
- SOMA Task - The according SOMA Label
- CRAM Comment - @gaya- and her CRAM team can comment on  e.g. the questions from the SOMA team
- SOMA Comments - SOMA team can add questions and comments which are related to the CRAM Action
- Status - The current status, e.g if own defined labels are used, currently in discussion or done.

CRAM Action | SOMA Task | CRAM Comment | SOMA Comment | STATUS
------------ | ------------- | ------------- | ------------- | -------------
Accessing | Accessing | | | Seba: I defined the SOMA Label by myself 
Closing |  Closing ||| DONE
Delivering | Delivering ||| Seba: I defined the SOMA Label by myself 
Detecting |  Detecting ||| Seba: I defined the SOMA Label by myself 
Fetching | Fetching ||| Seba: I defined the SOMA Label by myself 
GetActionGrasps | GetActionGrasps ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetActionGripperOpening  | GetActionGripperOpening ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetActionGrippingEffort | GetActionGrippingEffort ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetActionTrajectory | GetActionTrajectory  ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetContainerClosingDistance | GetContainerClosingDistance  ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetContainerOpeningDistance | GetContainerOpeningDistance  ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
GetLocationPoses | GetLocationPoses  ||| Seba: Is a reasoning task  and I defined the SOMA Label by myself
Grasping | Grasping ||| DONE
Lifting | Lifting ||| DONE
looking | LookingAt ||| DONE
searching | LookingFor ||| DONE
putting | Lowering ||| DONE
MonitoringJointState | MonitoringJointState ||| Seba: I defined the SOMA Label by myself 
Going | MovingTo ||| DONE
[...]

PR #159, opened 2020-11-23
A few more tasks added to match CRAM's repertoire

mpomarlan:
Partially addresses #158 and what's left out (GetX tasks, where X is a parameter) will be defined in cram_knowledge.

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -263,6 +263,7 @@ For Manipulating, it is the movement of the hand(s) and the change in functional
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Catching">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Actuating"/>
         <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#PickingUp"/>
+        <rdfs:comment>A task by which an Agent stops a moving object and gains kinematic control over it, usually by grasping.</rdfs:comment>
     </owl:Class>
     
 
@@ -527,6 +528,7 @@ Deduction is often explained as starting from the &quot;general&quot; (some prop
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Holding">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Manipulating"/>
+        <rdfs:comment>A task by which an Agent keeps an object over which it has kinematic control, typically via grasping, at some specified pose.</rdfs:comment>
     </owl:Class>
     
 
@@ -575,6 +577,7 @@ Induction is often described as a move from many &quot;specifics&quot; (swan A i
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#LookingFor">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Perceiving"/>
+        <rdfs:comment>A task by which an Agent uses its perception apparatus to check for the presence of an object in some specified area.</rdfs:comment>
     </owl:Class>
     
 
@@ -680,6 +683,7 @@ One could argue Mental actions are all Physical actions, because anything the Ag
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionEvaluationTopic">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionTopic"/>
+        <rdfs:comment>A topic used while an Agent describes its own cognitive processes and acions to evaluate them according to some metric.</rdfs:comment>
     </owl:Class>
     
 
@@ -688,6 +692,7 @@ One could argue Mental actions are all Physical actions, because anything the Ag
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionMemoryTopic">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionTopic"/>
+        <rdfs:comment>A topic used while an Agent describes its own cognitive processes and actions, and which covers descriptions of what memories are involved in them.</rdfs:comment>
     </owl:Class>
     
 
@@ -696,6 +701,7 @@ One could argue Mental actions are all Physical actions, because anything the Ag
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionPlanningTopic">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#MetaCognitionTopic"/>
+        <rdfs:comment>A topic used while an Agent describes the planning it does for its own cognitive processes and actions.</rdfs:comment>
     </owl:Class>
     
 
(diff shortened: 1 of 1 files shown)


daniel86 (changes requested):
@mpomarlan there are some warnings reported. please have a look at them! you can find them best in the DIFF page of github https://github.com/ease-crc/soma/pull/159/files

mpomarlan:
The warnings don't seem associated to the newly added tasks, but this might be a good time to tackle some anyway; skimming, I see several that just require comments to be added about some of the tasks. Ok I'll do this soon.

owl/SOMA-ACT.owl
```diff
@@ -379,6 +407,18 @@ Deduction is often explained as starting from the &quot;general&quot; (some prop
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#Delivering -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Delivering">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Actuating"/>
+        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Closing"/>
+        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Fetching"/>
+        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Opening"/>
```
daniel86, line 416: Is this a complete list? I wonder because above in Closing there are more disjoint classes e.g. Squeezing.
This is because squeezing is not necessary an actuation? Like when I sit on a pillow it would be squeezing but no actuation?

owl/SOMA-ACT.owl
```diff
@@ -528,6 +587,26 @@ Induction is often described as a move from many &quot;specifics&quot; (swan A i
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#MakingInteriorAccessible -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#MakingInteriorAccessible">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ModifyingPhysicalObject"/>
+        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#MakingInteriorInaccessible"/>
+        <rdfs:comment>A task in which an Agent will affect a container such that the container's interior is accessible for furhter tasks. The MakingInteriorAccessible task will include locating the container, navigating to it, and manipulating it appropriately to enact this state change.</rdfs:comment>
```
daniel86, line 595: How is this different from an Opening task? I guess the idea is to just classify the manipulation itself as opening?

My concern is that this seems very specific and would rather belong in the Plan section where we can represent concrete workflows of steps following each other as you have done in the comment. in my intuition the workflow described in the comment is just one of many ways of executing an opening task. One could say that the activity that executes this task has several motion phases, and in the case of the workflow above these would be navigation, approaching, grasping, etc. but we would only use one task label for all the events (however different motion labels)

daniel86 (changes requested):

owl/SOMA-ACT.owl
mpomarlan: No, initially my idea was that some of these tasks might coincide (e.g., a Lifting might be Delivering). However, on second thought maybe it's better not to have this. If needed, Tasks that are disjoint can nonetheless still classify the same Action if we needed them to. So I made it such that the disjointWith lists are against all other concepts in Actuating.

owl/SOMA-ACT.owl
mpomarlan: The rationale behind this being a task is the following:

there exists a CRAM task called Accessing (find it in Seba's list). This is a complex task including navigating to a container and includes Opening as a subtask. To the CRAM task, we now have the MakeInteriorAccessible task as a correspondent.

Naturally, this suggested the need for its opposite, and hence MakeInteriorInaccessible was born. I wouldn't make a distinction between "Plan" and "Task" since these are, as we discussed before, not very well-defined distinctions. You can always split a task into several subtasks; conversely, maybe the level of granularity at which you describe an action is in terms of goals such as making container interiors (in)accessible, without caring for the specifics of manipulation of how to render them thus.

mpomarlan: I mean, I wouldn't think it is a good idea to have certain CRAM "tasks" correspond to "Tasks" and others to "Plans" just on the basis of some increased complexity in terms of "subtasking" relations. Of course, Plans and Tasks are different ;)

owl/SOMA-ACT.owl
daniel86: Well, being disjoint does not mean that it is not allowed to classify the same event, right? Tasks do not really coincidence at all, only the events that execute them

owl/SOMA-ACT.owl
daniel86: > there exists a CRAM task called Accessing (find it in Seba's list). This is a complex task including navigating to a container and includes Opening as a subtask. To the CRAM task, we now have the MakeInteriorAccessible task as a correspondent.

But the act of "accessing" is not the same as "making accessible". In my opinion only the "opening" subtask corresponds to "making accessible"

daniel86: > The rationale behind this being a task is the following:
> 
> there exists a CRAM task called Accessing (find it in Seba's list). This is a complex task including navigating to a container and includes Opening as a subtask. To the CRAM task, we now have the MakeInteriorAccessible task as a correspondent.
> 
> Naturally, this suggested the need for its opposite, and hence MakeInteriorInaccessible was born. I wouldn't make a distinction between "Plan" and "Task" since these are, as we discussed before, not very well-defined distinctions. You can always split a task into several subtasks; conversely, maybe the level of granularity at which you describe an action is in terms of goals such as making container interiors (in)accessible, without caring for the specifics of manipulation of how to render them thus.

You haven't really answered my question how this is different from opening.
So my suggestion is still to stick to Opening and Closing, and these can be subtasks of accessing according to some plan where e.g. something is accessed from a container that is opened first, and closed again afterwards

daniel86: > I mean, I wouldn't think it is a good idea to have certain CRAM "tasks" correspond to "Tasks" and others to "Plans" just on the basis of some increased complexity in terms of "subtasking" relations. Of course, Plans and Tasks are different ;)

Well I was mostly referring to the comment wrt "plan" because I think a task should not be defined by the workflow that executes them but by the roles objects play and by the goal. however it should leave open how the goal is achieved, this is defined in plans

daniel86 (changes requested):
discussion is ongoing

also there are more warnings. unfortunately github has a limit of how many of them are displayed. to see all of them at once you need to download the data file from CI

owl/SOMA-ACT.owl
mpomarlan: Exactly. So the tasks can be of disjoint classes, and still classify the same event if that same event enacts two tasks somehow.

owl/SOMA-ACT.owl
mpomarlan: The CRAM task Accessing (to which MakeInteriorAccessible is intended to correspond) does however include both getting to the container and the Opening manipulation task. That's the difference.

If you think this partitioning is somehow too CRAM-specific, I'm happy to move MakeInterior(In)Accessible to cram_knowledge, but I don't think it really is as CRAM-specific as it seems. The problem rather comes from the colloquial use of "open"; it can refer to either the full action of going to the container and opening it, or to just the manipulation to open it once there. CRAM makes a distinction between the two, and it's a distinction worth reflecting in SOMA.

mpomarlan:
None of the CI warnings I see are related to the PR though. I can fix the "no human readable annotations" for SOMA-ACT concepts, but I will not touch those of other files.

mpomarlan:
An observation while going through SOMA-ACT and adding comments: some concepts that already have comments (Transporting, Holding) are included in the warning about missing human readable annotations, even though they have comments defined.

Again, I will go through the concepts in SOMA-ACT and add comments. I will not make it a point of this PR to fix those warning issues. This is better handled by another PR.

daniel86 (approved):
There are still some robot-specific definitions. I might suggest to move them to knowrob.owl or cram.owl at a later point

PR #163, opened 2020-12-14
Head and finger concepts are added to SOMA

AbhijitVyas:


mpomarlan (approved):

mpomarlan:
A couple notes on the quick approval:

-- I'm still puzzled that the Head concept isn't already present in SOMA; I remember creating something like this at some point, but perhaps it fell out and in any case I can't find it in SOMA-OBJ where it should have been.

-- we will have to return to establishing a hierarchy of functional parts of bodies for agents such as humans and robots, but this will remain something to do next year. For now, in the interest of having concepts that enable us to use NEEMs coming from the various groups, I suggest we do a quick patch-in of the concepts we need, such as those added via this PR. I'll open an issue we can review next year on the part hierarchy.

owl/SOMA-OBJ.owl
```diff
@@ -1704,6 +1704,26 @@ The entity playing an AgentRole is endowed with sentience and the capacity to de
         <rdfs:comment>A limb used to reach for objects.</rdfs:comment>
     </owl:Class>
     
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#Head -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Head">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#FunctionalPart"/>
+        <rdfs:comment>A functional part of body responsible for carrying high bandwidth sensors ie. camera.</rdfs:comment>
```
daniel86, line 1711: of *the* body

daniel86: *, i.e. ,*

daniel86 (approved):

PR #167, opened 2021-01-25
Added comments to some SOMA entities

mpomarlan:
NOTE: PR is not done yet. The purpose of opening it was to have a discussion board here to clarify some things (see next message).

Currently done: "most" of SOMA.owl, except for a few entities that need clarification.

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -63,6 +63,7 @@
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasConstituent"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Action"/>
         <rdfs:range rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Action"/>
+        <rdfs:comment>A relation from an Action to a component Action.</rdfs:comment>
         <rdfs:label>has action</rdfs:label>
     </owl:ObjectProperty>
     
@@ -74,6 +75,7 @@
         <rdfs:subPropertyOf rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRegion"/>
         <rdfs:domain rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Action"/>
         <rdfs:range rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#ExecutionStateRegion"/>
+        <rdfs:comment>A relation from an Action to its execution state.</rdfs:comment>
         <rdfs:label>has execution state</rdfs:label>
     </owl:ObjectProperty>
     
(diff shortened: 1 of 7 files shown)


mpomarlan:
Some entities I didn't touch because I don't understand how they would be used:

assistedBy
hasGoal (is this goal possessed or goal served?)
hasInputParameter (should domain be Concept? what does this mean?)
hasOutputParameter (-"-)
hasStatus: (how will this be used? or maybe redefine domain and range?)
isAnsweredBy: (-"-)
isAskedBy: (-"-)
isInputParameterFor: (-"-)
isOutputParameterFor: (-"-)
isReificationOf: what will this be?

Also a couple stragglers:
SOMA-OBJ#FunctionalPart: the prefix is wrong, but also this seems defined in another file. Perhaps some refactoring of the definition necessary, or moving what depends on it in SOMA.owl to the appropriate other file (SOMA-OBJ).

State: this is defined in SOMA-State.owl but used already in SOMA. As above, some refactoring seems in order.

mpomarlan:
Done some more commenting in SOMA.owl. A couple of issues appearing:

1) FunctionalPart is defined in SOMA-OBJ, but PhysicalEffector which needs it is already defined in SOMA.owl. I would suggest we either move FunctionalPart to SOMA.owl, or PhysicalEffector (and involvesPhysicalEffector, isPhysicalEffectorInvolvedIn) into SOMA-OBJ.owl

2) State is defined in SOMA-State, but used in the definition of hasPhase in SOMA.owl. As in the case of FunctionalPart, lets try to either define hasPhase in SOMA-State too or move State to SOMA.owl.

3) There are many entities defined using the SOMA-OBJ.owl prefix in SOMA-OBJ.owl. Does this need correcting?

4) Plausible concepts that can restrict the domain of isAskedBy and isAnsweredBy appear in SOMA-SAY.owl. Can we then move these relations to that file?

daniel86:
> Done some more commenting in SOMA.owl. A couple of issues appearing:
> 
>     1. FunctionalPart is defined in SOMA-OBJ, but PhysicalEffector which needs it is already defined in SOMA.owl. I would suggest we either move FunctionalPart to SOMA.owl, or PhysicalEffector (and involvesPhysicalEffector, isPhysicalEffectorInvolvedIn) into SOMA-OBJ.owl
> 
>     2. State is defined in SOMA-State, but used in the definition of hasPhase in SOMA.owl. As in the case of FunctionalPart, lets try to either define hasPhase in SOMA-State too or move State to SOMA.owl.
> 
>     3. There are many entities defined using the SOMA-OBJ.owl prefix in SOMA-OBJ.owl. Does this need correcting?
> 
>     4. Plausible concepts that can restrict the domain of isAskedBy and isAnsweredBy appear in SOMA-SAY.owl. Can we then move these relations to that file?

1. I prefer moving FunctionalPart to SOMA-OBJ
2. hasPhase range includes process or? so I would say it needs to go to SOMA.owl
3. it wouldn't hurt. But I think the IRIs are all rewritten in the composed version to use a single namespace. So it would not make a difference.
4. ok

AbhijitVyas (approved):

daniel86 (approved):

PR #173, opened 2021-02-15
information about triple data as JSON format is added to NEEM acquisition

AbhijitVyas:


docs/NEEM-handbook/content/acquisition/acquisition.tex
```diff
@@ -53,6 +53,21 @@ \section{Data Structure}
 	\item[\textbf{triples}] The triples collection contains all triples which were asserted into the knowledge base during run time.
 \end{description}
 
+\subsection{Triple data as JSON object}
+	One can provide triple data into JSON format logged in their experiment, where triples are defined in standard subject, predicate and object format. Subject and objects can be defined with valid IRI, which is pointing to concepts or individuals defined by ontology. An individual name is the combination of concept name and random string connected with an underscore. An object can be also other data types, such as a number, string, boolean, etc. Whereas, the predicate is defined by IRI pointing to property concepts mentioned in  ontology. One can also provide additional time scope fields `since` and `until` to indicate that the given triple is valid for the given time scope. An example of such JSON object is given below:
```
daniel86, line 57: Better: "Triple data can also be provided in form of ..."
"One can .." seems not formal enough to me.

daniel86, line 57: "logged in *their* experiment" is invalid grammar.
I would suggest to remove "logged in their experiment"

daniel86, line 57: "where triples are defined in standard subject, predicate and object format"
what is the "standard format", I can not get the meaning of this sentence

daniel86, line 57: "Subject and objects" singular/plural mix-up
either both singular or both plural

daniel86, line 57: "can be defined with valid IRI"
what is a "valid" IRI, what does "defines" mean here?

daniel86, line 57: "defined by ontology" bad grammar. *an* is missing, or say "the SOMA ontology"
I would also write "defined in"

daniel86, line 57: "combination of **a** concept name and **a** random string"

daniel86, line 57: "An object can be also other data types" sounds odd.
If it is a number, then it is not an object!
So you rather say that triples can not only link subject with object, but also can link subject with data value which is represented using one of the base types string, bool, number

daniel86, line 57: "the predicate is defined by IRI"
first, again **an** is missing
second, the predicate is not **defined** by the IRI, it is **named** by the IRI.
Defining is something different

daniel86, line 57: "An example of such **an** JSON object"

docs/NEEM-handbook/content/acquisition/acquisition.tex
```diff
@@ -53,6 +53,21 @@ \section{Data Structure}
...
+   "s": "http://www.ease-crc.org/ont/SOMA.owl#Salad_PMRVYPJH",
+   "p": "http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasRole",
+   "o": "http://www.ease-crc.org/ont/SOMA.owl#Patient_YGJUVNDR",
+   "since": 27.739,
+   "until": 29.075
+ }
+]
+\end{lstlisting}
+
+It is important to note that, this is an intermediate data format which is not stored into database but is useful for providing data into agreed common format which can be accessed and inferred into knowledge-base.
```
daniel86, line 70: "which is not stored into database but is useful for providing data into agreed common format which can be accessed and inferred into knowledge-base." this seems odd
-->
"which is not equivalent with how the NEEM narrative is actually stored in databases.
The format described here rather serves as an easy-to-use interchange format."

daniel86 (changes requested):

docs/NEEM-handbook/content/acquisition/acquisition.tex
AbhijitVyas: S, P, O is triple standard format.

AbhijitVyas: valid IRI which is pointing to concepts or individuals defined in ontology

AbhijitVyas: I can remove word "valid" in order to avoid further confusion.

daniel86 (approved):

Issue #170, opened 2021-02-04
Inconsistency in LocatumRole and RelatumRole


shoeffner:
LocatumRole is defined as SubClass Of:
- classifies only PhysicalObject
- Role
- SpatialRelationRole

RelatumRole is defined as SubClass Of:
- classifies only PhysicalObject
- SpatialRelationRole

As the two are used in similar context, I think they either should both be direct subclasses of `Role`, or none of them should be. What do you think?

daniel86:
@mpomarlan I would assume you created these definitions?

daniel86:
should be fixed in #180

PR #180, opened 2021-03-11
Removed def. of LocatumRole from SOMA-SAY since it is in SOMA-OBJ alr…

mpomarlan:
…eady.


mpomarlan:
Addresses issue #170.

The problem was not an inconsistency but a redundancy -- for some reason LocatumRole was defined both in SOMA-SAY and SOMA-OBJ. It is now defined only in SOMA-OBJ, which is imported by SOMA-SAY.

daniel86 (approved):

PR #181, opened 2021-03-12
new concept called Cup is added to SOMA-HOME

AbhijitVyas:


mpomarlan (commented):
The change seems mergeable as far as I'm concerned, but if we add Cup, perhaps we should also add the dispositions a cup would have?

AbhijitVyas:
> The change seems mergeable as far as I'm concerned, but if we add Cup, perhaps we should also add the dispositions a cup would have?

Thanks for the suggestion, the main two types of dispositions we have discussed were: Containment and Grasp-ability. I took a look at the Container class which has Containment as disposition, and for the later, I have added new Disposition type and at the moment assigned to 'designed handle'.

daniel86:
@AbhijitVyas  *Graspability* does not seem to belong into SOMA-HOME. you can grasp non home things

mpomarlan:
I have an idea that might explain this.

First, why the empty concepts are detected. I pulled the latest version of master, started Protege, and opened SOMA-SAY (it's a file that should include most others). I then ran the Hermit reasoner, and of course it showed FormalLanguage and FormalExpressions as empty concepts, which would then also render empty the properties that use them as domain or range. Here's an inference that Hermit does to explain why FormalLanguage is empty (square brackets annotate where the axiom comes from):

FormalLanguage subclassof Language [IOLite]
FormalLanguage subclassof hasPart only FormalExpression [IOLite]
Language subclassof Description [IOLite]
FormalExpression subclassof InformationObject [IOLite]
Description DisjointWith InformationObject [DUL]
hasPart inverseOf isPartOf [DUL]
Reflexive: isPartOf [DUL]

In words: FormalLanguage can only have FormalExpressions as parts. Since an object is always its own part (isPartOf is reflexive), this means a FormalLanguage has a FormalLanguage as a part. Some other subsumption and disjointness axioms then produce the result that FormalLanguage is an empty concept.

None of the above axioms comes from us. Look online, and with some ctrl+F in the respective files of DUL and IOLite you will find them.

But how come we didn't see this before?

I searched my harddrive for a local copy of DUL. Helpfully, it has a changelog in the beginning, which (for my local copy) stops at version 3.31.

Checking the online DUL reveals this interesting bit in the changelog:

In 3.34, extended mereological support for parthood, introducing hasPropertPart (transitive) as a middle property between hasPart (transitive and reflexive) and hasComponent (asymmetric). This solution uses then "reflexive reduction" and "transitive reduction" design patterns (they allow to grant property characteristics through the superproperties, but not in the subproperties). Technically, mereology axioms would require that also hasProperPart be asymmetric, however a direct subproperty of an OWL non-simple property (hasPart) cannot be also asymmetric, hence the approximation. 
Added a n-ary parthood class in order to suggest an alternative pattern for time- (and space-)indexed part relations. In order to ensure that property characteristics hold also with parthood n-ary, a property chain is introduced which infers a direct dul:partOf property for each parthood individual.
Added a dul:realizesSelfInformation propery in otder to enable local reflexivity ('Self') axioms for all information realizations.

In other words, the latest version of DUL is not consistent with IOLite. We should notify the relevant people.

daniel86:
@mpomarlan thanks for figuring this out!
Yes we could contact the maintainer of IOLite, I assume. But the ontology meta block does not hold this information.

Also, we then need to fix the DUL version everywhere to 32:
http://www.ontologydesignpatterns.org/ont/dul/DUL_v32.owl

Looking at the versions available, there is one listed as "v34pre" which was changed some days ago. Probably this is the trouble maker, so maybe v33 would work too.

mpomarlan (approved):

PR #186, opened 2021-04-28
Pass DUL v32 and IOLite explicitly to hermit.

### LINKED ISSUE #185: Latest DUL.owl (v34) does not work together with IOLite
DUL version v34pre is incompatible with IOLite.
SOMA applications need to use an earlier version to avoid inconsistencies.
This is the reason CI is reporting errors at the moment.

This version could be used:
http://www.ontologydesignpatterns.org/ont/dul/DUL_v32.owl

shoeffner:
Closes #185.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -89,10 +89,16 @@ jobs:
           name: SOMA OWL
           path: ./build
 
+      - name: Download DUL v32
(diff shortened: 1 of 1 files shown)


shoeffner:
I still think @mpomarlan's idea of adapting SOMA (#185) might be the correct way to move on in the future, but for the time being, this change should allow building SOMA correctly.

daniel86 (approved):

PR #189, opened 2021-06-10
spoon concept added

AbhijitVyas:


daniel86 (approved):
Looks good. Would be nice though if we could include some disposition axioms in the future

PR #191, opened 2021-06-16
extra ease-tsd concepts

Mome:
This are the extra concepts (mostly objects), I used for the NEEMs of the review meeting. I put them in a new file, because I did not want to mess with your files.
Can you review them and give me feedback?


owl/SOMA-HOME.owl
```diff
@@ -593,10 +593,299 @@
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Shaker">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#DesignedContainer"/>
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/SOMA.owl#SidePlate -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#SidePlate">
```
AbhijitVyas, line 811: It is possible if we can stick to some sort of hierarchy here for plates? We already have dinner plate then can we stick to such classification of plates where we attached the function they should serve in their name? I have never encounter term called "SidePlate". Maybe this can help? I see "BreadAndButterPlate" name suits best to our description here.
https://www.tableking.com.au/types-of-tableware-plates/

AbhijitVyas:
@Mome Hello, I have reviewed your pull request, I wanted to update it along with our disposition related information. So I have taken all of the objects that you have defined it here and created new pull request. So I guess we can close this one now.

owl/SOMA-HOME.owl
Mome: So I measuerd the plate and it is 20cm in diameter. We actually just distinguish small and large plate.
So I think from your list **BreakfastPlate** would fit best, and I would be fine with that.

AbhijitVyas: okay then. Now, we can add ease-tsd concepts to SOMA-HOME.

PR #195, opened 2021-10-14
Pre/postconditions, manifestations, preventions. And some cleanup.

mpomarlan:
(I will explain the cleanups, and a surprise in the diff, in the next comment.)

The approach I propose is the following: we have the Events to refer to things that actually happen. Events are the initial building bricks of our logs. We can add semantics (aka meaning aka interpretations) to Events via other types of concepts. In particular, we would now say that Events manifest Situations, which themselves satisfy some Descriptions from whatever theories, planning domain descriptions, etc. we have available.

The reason for this is that Situations don't have to manifest in Events, so we can speak of prevented Situations, or expected but not occurring Situations. As an example, we might have an Action A, which is the manifestation of some Situation S, and we expect this Action to result in a particular outcome P. We model this by having a triple (S, hasPostcondition, P). 

On the other hand, the outcome we actually observe is W (manifested in some state of the world X). We would model this by having triples (S, hasTerminalSituation, W) and (W, manifestsIn, X).

Maybe W is compatible with P, or is P (i.e., what we observe is what we expect). We can model this by the triple (S, hasExpectedTerminalSituation, P).

Maybe W is incompatible with P. At this point, it would be nice to assert a triple (P, isA, NonmanifestedSituation). This way, later on when we bring in the full failrecont into SOMA, we can do reasoning about failure.

So to recap: Situations may connect to each other via

hasInitialSituation, hasTerminalSituation (these can be used by SituationTransition, itself a kind of Situation): these assert that a manifested SituationTransition follows/precedes some other manifested Situations. This is to connect our interpretations of events that actually happened.

hasPrecondition: if (S, hasPrecondition, Z) then Z must manifest before S can manifest.
hasPostcondition: if (S, hasPostcondition, Z) then we expect Z to manifest after S manifests. Note that this may not be the case: the manifestation of S may fail to produce a manifestation of Z.

prevents: a Situation that manifests in an Event may prevent another from manifesting.

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -438,6 +438,7 @@ Deduction is often explained as starting from the &quot;general&quot; (some prop
 
     <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA.owl#Delivering">
         <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Actuating"/>
+        <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Closing"/>
         <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Fetching"/>
         <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Lifting"/>
         <owl:disjointWith rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Opening"/>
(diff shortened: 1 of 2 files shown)


mpomarlan:
Update on the deleted "DisjointWith" assertions.

This seems to be Protege trying to be helpful and reduce the file size. Notice, for example, that "Closing" is asserted to be DisjointWith Delivering; on the other hand, the disjointness assertion between Delivering and Closing is deleted. Indeed, the second assertion is redundant.

mpomarlan:
Updated SOMA-ACT: restored redundant disjointness axioms, as discussed in our onto meeting yesterday.

PR #209, opened 2022-05-06
Added concepts: Oven, Kitchen, TrashContainer

kaviyachandran:
These concepts are added to be used in iai kitchen owl file


owl/SOMA-HOME.owl
```diff
@@ -184,6 +184,15 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-HOME.owl#Kitchen -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-HOME.owl#Kitchen">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Room"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A designated room to perform the task of cooking</rdfs:comment>
```
daniel86, line 191: What type of dispositions does it have? I think we should explicate that ;)
Would be actually nice for all concepts here.

daniel86 (changes requested):

owl/SOMA-HOME.owl
daniel86: any progress?

daniel86: @kaviyachandran ?

kaviyachandran: Hi Daniel, I have updated the PR with dispositions for the objects

daniel86: @kaviyachandran the pull request still has the status "requested changes must be addressed". you need to "re-request review" to trigger another review.
But it seems conflicts popped up that need to be solved first.

kaviyachandran: Hi Daniel, I fixed the conflicts.

mpomarlan: The containment disposition should go to the room concept itself. What would be a disposition of a kitchen, specifically? This question @daniel86 too.

E.g., both kitchens and bathrooms tend to be what in Germany is called nassbereich. I.e., these rooms should be able to handle a larger level of moisture than other rooms of a house (and there are restrictions on what construction materials to use for the wall and floor coverings as a result).

There is of course also the fact that the kitchen is the room intended for cooking :) I'm not sure about whether this is a disposition of the room itself or a result of various devices and tools being typically located there.

owl/SOMA-HOME.owl
```diff
@@ -220,6 +235,38 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#DesignedContainer"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#hasDisposition"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Containment"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#hasDisposition"/>
```
mpomarlan, line 251: I don't think this is the way to go to assert the oven has a disposition to expose its contents to heat. It is not clear that a bakedgood will be a trigger only for such a disposition, nor is it true that this disposition requires a baked good to be involved in the baking process. Perhaps I want to use the oven to melt butter, for example, or age a piece of paper for a school project in history class.

If I recall, there is a Tempering disposition. It, or some appropriate subclass of it, should be used.

owl/SOMA-HOME.owl
```diff
@@ -385,6 +432,21 @@
     
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-HOME.owl#TrashContainer -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-HOME.owl#TrashContainer">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#DesignedContainer"/>
```
mpomarlan, line 438: I would put the containment disposition to the designedcontainer class instead.

Are there any trash-container specific dispositions? This question also @daniel86 

For example, trash containers often -- but not always -- have some sort of lid or other covering mechanism built in. This is because they are intended to prevent smells from their contents from readily propagating and/or limit access of insects to the trash.

mpomarlan (changes requested):

mpomarlan:
We discussed in our previous onto meeting that this PR should be resolved quickly, unfortunately this falls inside Kaviya's holiday.

So, a somewhat exceptional procedure:
I will close this pull request, however update my own fork of soma with its contents. I will implement the changes I requested, open a new PR, and ask for a new review.

Issue #210, opened 2022-05-11
Hosting DUL ourselves


mrnolte:
Hi there,
It occurred repeatedly that DUL was not reachable via its IRI. Can we clone DUL and host it ourselves? This way we could also adjust unwanted definitions or concepts in DUL.

mpomarlan:
I don't mind hosting it -- though so far I haven't had problems accessing it.

I would advise against adjusting it. To the extent that we wish to claim founding on DUL helps us connect to other ontologies of similar foundation, we should obey it as is and avoid schism.

mpomarlan:
Further comment, just to make things "clearer": I'm quite fond of occasionally attempting to rewrite DUL in a lower formalism, e.g. EL++. So far though those would be personal hobby projects. A more serious decision would need a bit of politics as to how we position SOMA relative to what else is there in knowledge engineering.

mrnolte:
Well, try aceessing DUL right now: http://www.ontologydesignpatterns.org/ont/dul/DUL.owl

mpomarlan:
Oops so it is. The whole ontologydesignpatterns.org is down.

mrnolte:
Yep. But I see your point not to change DUL; and while I agree with that, we sometimes understand and use Concepts and Relations not according to their original purpose. It might make things clearer. if we update the comments in DUL for those cases.

mpomarlan:
Yes, that sounds quite useful.

sasjonge:
This was implemented by

https://github.com/ease-crc/soma/pull/212

Issue #202, opened 2022-03-01
SOMA import structure needs to be fixed


mrnolte:
Currently, the situation is as follows:
The repository contains 11 `.owl` files.
`SOMA.owl` only imports DUL, the other files (in some way or another) import SOMA, but located at [http://www.ease-crc.org/ont/SOMA.owl](http://www.ease-crc.org/ont/SOMA.owl).

The latter, however, is _not_ the same as `SOMA.owl`, as it contains additional classes, e.g., `Disposition`. To my understanding, it is created automatically as the union of all / a subset of the `.owl` files in the git.

This is problematic when developing or changing the `.owl` files in the git, because the online version does not change dynamically, which is good for distribution, but bad for development as the `.owl` files would always import their old version via the online import. Changes will not be reflected elsewhere.

Furthermore, it is hard to find the sources of specific classes or other entities as in protege, due to the online import it looks always as if the classes stem from SOMA.

Finally, it is confusing that all ontologies import each other as they each import the collapsed online version. The separation of the sub-ontologies in different files is therefore useless.

Proposal:
The git `.owl` files should not import the collapsed version, but instead they should import the `SOMA.owl` in the git. It would be great if the use relative paths for that. Otherwise maybe write a script to host them under `localhost` or something.

mrnolte:
As a workaround: In the [owl](../tree/master/owl) directory, Protégé automatically creates a file called `catalog-v001.xml` when opening any ontology. This lets us redirect the import from the URL `http://www.ease-crc.org/ont/SOMA.owl` to a relative path `soma.owl`. Just add something like the foillowing line in the group-tag:
```xml
<uri id="Automatically generated entry, Timestamp=1649666788715" name="http://www.ease-crc.org/ont/SOMA.owl" uri="SOMA.owl"/>
```
Protégé will then resolve SOMA as required for development.

mrnolte:
Has been discussed in todays meeting as should-be-done.

PR #211, opened 2022-05-13
Protégé work environment setup

mrnolte:
Three notable changes:

1. In protégé, you can specify where to load ontologies using a catalog-file. This file was previously ignored, as protégé automatically generates it, but it is very useful for development (and will be ignored by all other applications). Therefore I removed it from the `.gitignore`. I can see the point to not put that file at the master branch, and to put it at a development branch. However, the primary way to distribute SOMA is not via this repository but via its IRI. Also, we already include multiple development related files, e.g., in the `.github` directory.
2. Redirection of all SOMa-OWL to their copy within the `owl` directory using the catalog-file mentioned above. This solves #202.
3. I added DUL.owl, as that ontologies IRI is notoriously unreliable and has been unavailable repeatedly for extended periods of time. I redirected to that file in the catalog. This also allows for complete out-of-the-box offline development and solves #210.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@
 *.zip
 
 # ignore generated files
-*catalog-v001.xml
 *EASE-UGLY.owl
 SOMA.owl
 build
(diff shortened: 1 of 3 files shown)


mrnolte:
Maybe the deplyment should also make use of the catalog-file, such that hermit etc. work without loading DUL?

sasjonge:
I'm not sure why this PR shows conflicts. Maybe you could create a new one and close this one, only with the catalog file.

PR #212, opened 2022-05-13
Moved DUL to a local server

sasjonge:
I added DUL and changed the deployment workflow, so that DUL should be accessible over our server.

For the moment I removed left over IOLite imports. We have to change references in some files.


mrnolte:
Can you add an Issue for the changes regarding IO Light?

Issue #226, opened 2022-05-19
SOMA NEEM


mrnolte:
Create an Ontology SOMA NEEM and move relevant stuff there, e.g., KynoDinamicData. Has been discussed in the meeting as should-be-done.

PR #228, opened 2022-05-20
Created SOMA-NEEM

mrnolte:
As discussed yesterday, I created SOMA-NEEM and moved three specific concepts there.


mrnolte:
Fix for #226

sasjonge (commented):
There are two locations where you have to potentially add the new owl file:

- In the uglify script:
https://github.com/ease-crc/soma/blob/caa558be8b9d2a5af5ee803a4e2144a0b504daac/prolog/uglify.pl#L65
- And potentially here: https://github.com/ease-crc/soma/blob/master/scripts/ecv.sh

I'm not sure about the last one, maybe @mpomarlan knows if this script is still important.

mrnolte:
Well that is not intuitive at all. Why don't these scripts just use all ontologies in the owl directory?

daniel86 (approved):

Issue #219, opened 2022-05-13
Use Konclude instead of Hermit for CI


mrnolte:
Because Konclude is MUCH faster (up to 50.000 times)

sasjonge:
@mrnolte Can you provide a script like this for Konclude:

https://github.com/ease-crc/soma/blob/master/scripts/hermit_test.py

Then I could add it to the CI

sasjonge:
We could also solve this Issue in the same PR: 

https://github.com/ease-crc/soma/issues/206

mrnolte:
I unfortunately have no idea how to update the CI, although i can certainly write a java script that asks conclude for the subclasses etc

mrnolte:
Not javascript, a script written in Java. But i guess we can try to use konclude this way to

mrnolte:
However, i only used Konclude via OwlLink - maybe Mihai knows best using Konclude via Command Line?

mpomarlan:
I'm writing a version of this script for Konclude but I encounter a problem: Konclude doesn't seem to return an object property hierarchy.

The "classification" command (available for the command line) returns the class hierarchy only. The command line also allows sending an owllink request, but the GetSubObjectProperties and GetSubObjectPropertyHierarchy return "node not supported" messages.

On the other hand, maybe this isn't too big a limitation. Whenever we've had a problem with a property being empty, so far it was because its domain or its range was empty, and we would still catch that. I'll write the script and make a pull request later. We can always add stuff if we figure out something/Konclude adds new features.

PR #231, opened 2022-05-31
Added konclude_test.py

mpomarlan:
Part of addressing #219 

After the konclude_test.py script is accepted, the evaluation.yml file needs to be changed.

### DIFF
--- a/scripts/konclude_test.py
+++ b/scripts/konclude_test.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python
+import argparse
+import re
+import sys
+from pathlib import Path
+
+
+OWL_NOTHING = 'http://www.w3.org/2002/07/owl#Nothing'
+BOTTOM_OBJECT_PROPERTY = 'http://www.w3.org/2002/07/owl#bottomObjectProperty'
+
+## IRI = '<(.*?)>'
+
+## EQUIVALENT_CLASSES = r'^EquivalentClasses\( (.*) \)$'
+## OBJECT_INVERSE_OF = r'ObjectInverseOf\( ' + IRI + r' \)'
+## EQUIVALENT_OBJECT_PROPERTIES = r'^EquivalentObjectProperties\( (.*) \)$'
+
+def _parseHomebrew(fileText):
+    inEq = False
+    nullClasses = []
+    aux = []
+    iriPLen = len("<Class IRI=\"")
+    for l in fileText.splitlines():
+        l = l.strip()
+        if inEq:
+            if "</EquivalentClasses>" == l:
+                inEq = False
+                if OWL_NOTHING in aux:
+                    nullClasses = nullClasses + aux
+                aux = []
+            else:
+                if l.startswith("<Class IRI=\""):
+                    aux.append(l[iriPLen:-3])
+        else:
+            if "<EquivalentClasses>" == l:
+                inEq = True
+    nullClasses = list(set(nullClasses))
+    if OWL_NOTHING in nullClasses:
+        nullClasses.remove(OWL_NOTHING)
+    return nullClasses
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('konclude_output', help='Konclude output file', type=str)
+    parser.add_argument('-o', '--output', help='output file', default='konclude.html', type=str)
+    args = parser.parse_args()
+
+    nullClasses = _parseHomebrew(open(args.konclude_output).read())
+    results = {
+        'EquivalentClasses': [nullClasses],
+        'SubClassOf': [],
+        'EquivalentObjectProperties': [],
+        'SubObjectPropertyOf': [],
+        'Error': [],
+    }
+
+    output = format_output(results, format=args.output[args.output.rfind('.'):])
+    with open(args.output, 'w') as outfile:
+        _ = outfile.write(output)
+    print(format_output(results))
+
+    # check if there are problems, return non-zero exit code if so
+    if len(nullClasses) != 0:
+        sys.exit(1)
+
+
+def as_html(titles, results):
+    template = '<h2>{title}</h2>\n{issues}'
+    outer = '<section>\n<p><h3>Set</h3>{}</p>\n</section>\n'
+    outeritem = '</p><p><h3>Set</h3>'
+    inner = '<ul>\n<li>{}</li>\n</ul>\n'
+    inneritem = '</li>\n<li>'
+    sections = []
+    for k in sorted(results.keys()):
+        sets = results[k]
+        if not sets:
+            issues = '<p>No issues detected.</p>'
+        else:
+            inners = list(inner.format(inneritem.join(sorted(s))) for s in sets)
+            issues = outer.format(outeritem.join(inners))
+        section = template.format(title=titles[k], issues=issues)
+        sections.append(section)
+    sections = '\n'.join(sections)
+    return "<!DOCTYPE html>\n<html>\n\t<head>\n\t\t<title>Konclude results</title>\n\t\t<meta charset='utf-8' />\n\t</head>\n\t<body>\n\t\t<h1>Konclude results</h1>\n\t\t\t" + sections + "\n\t</body>\n</html>"
+
+
+def as_text(titles, results):
+    template = '\n## {title}\n\n{issues}'
+    outer = '* {}'
+    outeritem = '\n* '
+    inner = '- {}\n'
+    inneritem = '\n  - '
+    sections = []
+    for k in sorted(results.keys()):
+        sets = results[k]
+        if not sets:
+            issues = 'No issues detected.\n'
+        else:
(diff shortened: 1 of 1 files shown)


mpomarlan:
I've also updated the evaluation.yml file.

@sasjonge do you know how to write Docker files so that Konclude is also available?

Another couple of things to fix:

a) format. Konclude supports only OWL/XML and OWL functional syntax, not the RDF/XML that SOMA is written in. This needs an intermediate conversion step.
b) imports. These can be URIs to some web address and Konclude will find them just fine, however the same as above applies about the format. I would suggest here that we host DUL, and that we host an OWL functional syntax version of it. IOLite will be removed as an import if it hasn't been already so that's not important.

sasjonge:
I can have a look. I saw that konclude already has an existing docker image:

https://hub.docker.com/r/konclude/konclude

Maybe we can use that, or atleast their Dockerfile to create our own image.

sasjonge:
I tested using the konclude container and it seems to generally work:

https://github.com/sasjonge/soma/blob/cf97673dd1122b616906d0f05331018e0a7cccc4/.github/workflows/evaluation.yml#L79

But if fails on this step, because a action is missing:

https://github.com/sasjonge/soma/blob/cf97673dd1122b616906d0f05331018e0a7cccc4/.github/workflows/evaluation.yml#L99

Did you forget to push it?

mpomarlan:
Not so much forgot, as asking for assistance :) As far as I understand, the Konclude exec should be placed at the appropriate place by some kind of container setup. Don"t know how to do that.

mpomarlan:
So the the problem is this:

2022-07-05T16:01:22.3350599Z ## Classes equivalent to http://www.w3.org/2002/07/owl#Nothing
2022-07-05T16:01:22.3350791Z 
2022-07-05T16:01:22.3351106Z * - 
2022-07-05T16:01:22.3351223Z 
2022-07-05T16:01:22.3351228Z 

? I will have a look on my side.

sasjonge:
The other outputs are always:

No issues detected.

For  "## Classes equivalent to http://www.w3.org/2002/07/owl#Nothing" it is:

* -

sasjonge:
Can I merge my changes to the konclude-ci branch to extends this PR/Draft?
https://github.com/ease-crc/soma/pull/247

mpomarlan:
Pushed a commit that should fix this. The output will now be no issues detected if no classes are Nothing, and if there are unsatisfiable classes these will appear in a list.

NOTE: Konclude cannot test emptiness of properties. I may have mentioned this before, but just to make sure we're all reminded of this, here it is again. This shouldn't affect us too much as the typical way for a property to be empty is for either its domain or range to be empty. As long as we don't use intersections when specifying domains/ranges, we should be fine. If we use named concepts, or unions of named concepts, to specify domains/ranges, then the emptiness of the property would be caught by emptiness of the domain/range concepts.

sasjonge:
I merged my code into this branch over this PR https://github.com/ease-crc/soma/pull/247

mpomarlan:
Good. If the ci now works, this is ready for review.

mpomarlan:
"Yes". As long as I have a list of object properties r, then one thing to do is to create a "query owl file" containing concepts of the form "r some Thing" for each r, then do a classification.

The one thing I'm not sure how to do right now is to get that list of properties. One option would be to check the list of declarations in the SOMA ugly file, assuming that all used object properties are also declared.

mrnolte:
> The one thing I'm not sure how to do right now is to get that list of properties. One option would be to check the list of declarations in the SOMA ugly file, assuming that all used object properties are also declared.

It would not be a problem to generate such a list with the OWL API.

sasjonge:
> > The one thing I'm not sure how to do right now is to get that list of properties. One option would be to check the list of declarations in the SOMA ugly file, assuming that all used object properties are also declared.
> 
> It would not be a problem to generate such a list with the OWL API.

Is this necessary to approve this PR?

mpomarlan:
"Your" call -- for some value of "your".

As I said, what this means is the code as-is right now will not detect impossible properties unless they have impossible domains. This is maybe sufficient given that most of the domains we define are not intersections. On the other hand, we may have intersection domains, indeed the new versions of some of the GCIs were rewritten as such.

mrnolte:
If you tell me where to put such a list of properties, I will gladly export it so that you can then read it :)

mpomarlan:
Would the same folder housing the merged SOMA be ok? (I know I can access that one for Konclude input :) )

sasjonge:
I would rather recommend to have the path as an passed argument, so that we don't have to change the file if we change the locations in the pipeline later.

mpomarlan:
Sure, that also works.

So then something like script.py --propertyList path_to_property_list

?

sasjonge:
Yes, that looks okay.

mrnolte:
I now implemented exporting the properties. Just add a runtime argument propertyListPath to the java-ci stuff, e.g., as follows:
```cmd
spring-boot:run "-Dspring-boot.run.arguments=--versionInfo=current --propertyListPath=<path X>"
```
This will create 2 files for each of the "main" ontologies (with that I mean the minimal set of ontologies, so that all ontologies from within the owl directory are imported - currently, these are Allen, Home and Elan. Interesting path problem to identify those btw). For example, `<path X>/SOMA-HOME_objectProperties` and `<path X>/SOMA-HOME_dataProperties`

sasjonge:
@mrnolte I have a problem using the flag. I get a ¨Could not create dir" error in CI and also locally: https://github.com/sasjonge/soma/runs/7629898757?check_suite_focus=true#step:5:376

Also would it be possible to make the propertyExporter optional, so that the other CI steps don't need to execute it (and get the argument)?

mrnolte:
@sasjonge this should have fixed it, can you check? Also, if propertyListPath is not given, the CI will no skip it.
Maybe a dumb question, but do we really have to call the java CI twice? Is there no option to generate the list in the first run and use it in the second?

mpomarlan:
Updated the script. It will now optionally take an argument that gives a path to a file containing a list of object properties, each with its full IRI, one property per line.

The current line in the evaluation workflow reads:


python ./scripts/konclude_test.py -k ./.github/actions/Konclude -s ./build/SOMA-HOME.owl -o ./owl/konclude.html ./build/konclude.output

it should be 

python ./scripts/konclude_test.py -p <PATH TO PROPS LIST FILE> -k ./.github/actions/Konclude -s ./build/SOMA-HOME.owl -o ./owl/konclude.html ./build/konclude.output

Also please check if the paths to the Konclude binary (the -k arg) and the merged soma (-s) are correct.

sasjonge:
@mpomarlan I tried to change it in the CI, but if you now want to use the konclude binary from inside of your script it will make the setup more difficult, because I can't just use the konclude docker file. I'm wondering if we could change the system call in your script to a docker run call?

https://github.com/sasjonge/soma/blob/621354d12cd88cb4ec6a11e675e4e026932dd95f/scripts/konclude_test.py#L65

mpomarlan:
About SOMA_QUERY, the structure seems ok (I didn't notice Konclude complaining about parsing errors) and the content is intentional: when querying for properties, I only care about concepts of the form hasProperty some Thing (if hasProperty is equivalent to Bottom, then its domain is Nothing).

More interesting is that some properties have IRIs containing something else than SOMA or DUL, e.g. http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasChildLink.

This is coming from the property list file I imagine.

sasjonge:
> About SOMA_QUERY, the structure seems ok (I didn't notice Konclude complaining about parsing errors) and the content is intentional: when querying for properties, I only care about concepts of the form hasProperty some Thing (if hasProperty is equivalent to Bottom, then its domain is Nothing).
> 
> More interesting is that some properties have IRIs containing something else than SOMA or DUL, e.g. http://www.ease-crc.org/ont/SOMA-OBJ.owl#hasChildLink.
> 
> This is coming from the property list file I imagine.

@mrnolte Is that with this different IRI's a problem with the script?

sasjonge:
@mpomarlan 
I get an parsing error:

`XML parsing error at 1:1: 'Start tag expected.'.`

My newest run is this:

https://github.com/sasjonge/soma/actions/runs/3338066719/jobs/5525234865

mrnolte:
It is a problem that @daniel86 explained to me: Someone who worked on SOMA-OBJ did not configure protege correctly and created all kinds wrong iris. His old script renamed these. I wondered whether old neems use the SOMA-OBJ IRI and did not change my script to also rename those IRIs, but that should be no problem after my vacation.

mpomarlan:
The xml parsing error is not important as this is not an xml format, but the OWL functional syntax parsing error is. I think this is because the property list file has property IRIs with <> around the names whereas I assumed they would be without. If this is it then the fix is simple I'll have it done in a minute.

mpomarlan:
Should be done now.

mpomarlan:
Seems like a command line arg is wrong now, SOMA_HOM_objectProperties not found.

PR #235, opened 2022-06-14
Cut, Pour and Storage related dispositions 

AbhijitVyas:


owl/SOMA-OBJ.owl
```diff
@@ -1391,8 +1391,60 @@ In robotics, it is not uncommon to encounter shapes that are flipped compared to
...
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#CutObject -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#CutObject">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#AlteredObject"/>
+        <rdfs:comment>An object being cut down into pieces.</rdfs:comment>
+    </owl:Class>
+
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#Cutable -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Cutable">
```
kaviyachandran, line 1422: I defined [Cuttability](https://github.com/kaviyachandran/situational_awareness_example/blob/master/owl/situation_awareness_example.owl#L711) as a disposition property of the objects that take the role of CutObject

owl/SOMA-OBJ.owl
```diff
@@ -1391,8 +1391,60 @@ In robotics, it is not uncommon to encounter shapes that are flipped compared to
...
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>The disposition of an object (the barrier) which makes it able to be cut down (into pieces) usually by some other object with role as a Cutter. Such as a cucumber is cutable by a knife.</rdfs:comment>
+    </owl:Class>
     
 
+
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#Cutter -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Cutter">
```
kaviyachandran, line 1443: You could directly use CuttingTool

owl/SOMA-OBJ.owl
```diff
@@ -1682,6 +1704,28 @@ In robotics, it is not uncommon to encounter shapes that are flipped compared to
     </owl:Class>
 
 
+    <!-- http://www.ease-crc.org/ont/SOMA-OBJ.owl#Storage -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Storage">
```
kaviyachandran, line 1709: You can just use Containment

mpomarlan:
Having looked through this, I can summarize my review this way:

1) Kaviya's observation about overlaps with concepts from SOMA-HOME is a good one, and we should decide what to do about this redundancy.

2) I may need a refresher into how the affordance model of SOMA developped :P Here is my understanding of it so far, and correct me if I am wrong here. An affordance is a relation between dispositions of objects. If it can be said to describe a disposition, it describes all dispositions taking part in it. So, for example, an affordance about cutting describes both the disposition of an object to be cuttable, and the disposition of another object to cut others.

One of these dispositions is the "bearer", the other the "trigger". The difference is the following: the bearer is the one "acted upon", or more clearly, the bearer is the entity undergoing some significant process or change. The trigger may change too if you look at things in detail, but if you conceptualize something as a "trigger", you're saying you don't care about, as in, you abstract away from, any change in the trigger. Once a knife cuts a stick of butter it is still the same knife. Once the stick of butter gets cut, well you have to account for some different entities that have appeared etc.

There are complications using this heuristic of bearer vs trigger (what happens if the trigger also gets used up, such as when using a match to light a stove?) so here is another one, perhaps more robust: the bearer is the entity which undergoes some relevant change as a result of the affordance manifesting. Relevant means aligned or counter-aligned with someone's goals. As in, when I light the stove with a match, I do this because I intend the stove to be on fire; the fact that the match also gets burnt is incidental, it is neither an intention of mine nor do I intend the match to not be burnt.

To bring this round to the pull request here, I look at an axiom saying that CanCut affordsBearer CuttingTool with suspicion. CuttingTool is not a bearer role in an affordance about cutting, it is a trigger role. I see something similar happening for ShapedObject in an already accepted part of the ontology.

My suggestion -- can we discuss this in the onto meeting? I want to hear what the meaning of affordances is now, and why that changed.

owl/SOMA-OBJ.owl
```diff
@@ -1391,8 +1391,60 @@ In robotics, it is not uncommon to encounter shapes that are flipped compared to
...
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#affordsBearer"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Cutter"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#affordsTrigger"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#CutObject"/>
```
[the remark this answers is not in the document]
AbhijitVyas: As far as I see in SOMA-HOME.owl, a CuttingTool is of type DesignedTool (is a PhysicalObject) which has some taxonomy underneath such as Knife etc. My proposal of new concept called Cutter is not of type PO but rather a role which is assumed by PO such as a knife during a cutting event. I see dispositional axioms such as 'affordsBearer' and 'affordsTrigger' associated with Role and hence the concept of 'Cutter'.  We can for sure discuss this thing during over meeting today.

owl/SOMA-OBJ.owl
AbhijitVyas: See my comment above.

owl/SOMA-OBJ.owl
AbhijitVyas: yes but it is far more than a Containment. For example; a spoon contains a liquid but it can not store it. Hence a need for a new dispositional concept. :)

kaviyachandran: Yes, as Mihai mentioned, there needs to be a distinction between containment and storage

daniel86:
I removed myself and Sascha as reviewer. 4 reviews are not needed. @AbhijitVyas you should have enough input for updating the pull request now, or?

owl/SOMA-OBJ.owl
AbhijitVyas: @mpomarlan and @kaviyachandran , do we agree upon this discussion here? If yes I would like to resolve this conversation.

AbhijitVyas:
> Having looked through this, I can summarize my review this way:
> 
> 1. Kaviya's observation about overlaps with concepts from SOMA-HOME is a good one, and we should decide what to do about this redundancy.
> 2. I may need a refresher into how the affordance model of SOMA developped :P Here is my understanding of it so far, and correct me if I am wrong here. An affordance is a relation between dispositions of objects. If it can be said to describe a disposition, it describes all dispositions taking part in it. So, for example, an affordance about cutting describes both the disposition of an object to be cuttable, and the disposition of another object to cut others.
> 
> One of these dispositions is the "bearer", the other the "trigger". The difference is the following: the bearer is the one "acted upon", or more clearly, the bearer is the entity undergoing some significant process or change. The trigger may change too if you look at things in detail, but if you conceptualize something as a "trigger", you're saying you don't care about, as in, you abstract away from, any change in the trigger. Once a knife cuts a stick of butter it is still the same knife. Once the stick of butter gets cut, well you have to account for some different entities that have appeared etc.
> 
> There are complications using this heuristic of bearer vs trigger (what happens if the trigger also gets used up, such as when using a match to light a stove?) so here is another one, perhaps more robust: the bearer is the entity which undergoes some relevant change as a result of the affordance manifesting. Relevant means aligned or counter-aligned with someone's goals. As in, when I light the stove with a match, I do this because I intend the stove to be on fire; the fact that the match also gets burnt is incidental, it is neither an intention of mine nor do I intend the match to not be burnt.
> 
> To bring this round to the pull request here, I look at an axiom saying that CanCut affordsBearer CuttingTool with suspicion. CuttingTool is not a bearer role in an affordance about cutting, it is a trigger role. I see something similar happening for ShapedObject in an already accepted part of the ontology.
> 
> My suggestion -- can we discuss this in the onto meeting? I want to hear what the meaning of affordances is now, and why that changed.

We should discuss this PR in tomorrow's ontology meeting one last time in order to remove any confusion.

mpomarlan:
I see the branch has a few conflicts that make github unable to merge. I'll start the review when these are fixed or next week Wednesday, whichever comes later.

AbhijitVyas:
> I see the branch has a few conflicts that make github unable to merge. I'll start the review when these are fixed or next week Wednesday, whichever comes later.

Done. The conflicts are resolved.

mpomarlan:
Excellent!

owl/SOMA-OBJ.owl
```diff
@@ -1630,6 +1704,28 @@ In robotics, it is not uncommon to encounter shapes that are flipped compared to
...
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/SOMA-OBJ.owl#Storage">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#Enclosing"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#affordsTrigger"/>
+                <owl:allValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA-OBJ.owl#StoredObject"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>The disposition of an object (the container) to store other objects by enclosing them to prevent their free movement.</rdfs:comment>
```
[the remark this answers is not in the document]
kaviyachandran: > 

Does it make sense to say the objects which can be used to store has both dispositions containment and blockage?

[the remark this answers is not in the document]
AbhijitVyas: The same object can be used for multiple objectives and hence can poses many dispositions. But here the comment is leading to the direction of explaining how the container can be used to store things, and by doing so it can indirectly block free movements of things. However, new comment should not lead to such confusion.

mpomarlan (changes requested):
Almost done :)

I noticed PouredObjecct has a typo, and while minor it would be nice to fix before merging.

A little more significant is the problem with Pourable. This actually is the fault of the English language, as we tend to use the same syntactic forms for very different things: "Alice poured the milk into the bowl" vs "Alice poured the cup into the bowl".

What this means, first, is that the concept names are not suggestive. Pourable could as well refer to the liquid being poured, and not the container it comes from (and vice-versa for PouredObject). Maybe there is a way to have more suggestive concept names.

We can live with somewhat ambiguous concept names if the concept comments are hard to misinterpret. This actually affects the comments for both Pourable and PouredObject.

E.g., PouredObject: Alice pours the cup onto the table. So will the PouredObject refer to the cup, or to the liquid that was in it?

E.g., Pourable: the first sentence of the comment is ambiguous (see Alice above). The comment is further strange in that it seems to want to resolve this ambiguity, but does so in conflicting ways. On the one hand, we have "(the barrier)" in the first sentence which suggests we are talking about the source container. However the second sentence talks about the fluid.

I suggest that, since natural language use is a little ambiguous when pouring is concerned, it will be ok if the comments will be a little "strained" in terms of expression. E.g. something like the disposition of a fluid or substance which makes it possible to pour it out of a container and into or onto other objects, a role of some fluid or substance that is the patient of pouring task. Still not quite perfect as not only fluids are pourable (powder is pourable and I'm not sure it counts as a fluid).

owl/SOMA-OBJ.owl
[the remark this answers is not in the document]
kaviyachandran: I will just check with @mpomarlan once today in the meeting. Yes, we can add the disposition as an enabler of common everyday usage. We cant add all the possible dispositions. I did not mean it in that way.

AbhijitVyas:
> Almost done :)
> 
> I noticed PouredObjecct has a typo, and while minor it would be nice to fix before merging.
> 
> A little more significant is the problem with Pourable. This actually is the fault of the English language, as we tend to use the same syntactic forms for very different things: "Alice poured the milk into the bowl" vs "Alice poured the cup into the bowl".
> 
> What this means, first, is that the concept names are not suggestive. Pourable could as well refer to the liquid being poured, and not the container it comes from (and vice-versa for PouredObject). Maybe there is a way to have more suggestive concept names.
> 
> We can live with somewhat ambiguous concept names if the concept comments are hard to misinterpret. This actually affects the comments for both Pourable and PouredObject.
> 
> E.g., PouredObject: Alice pours the cup onto the table. So will the PouredObject refer to the cup, or to the liquid that was in it?
> 
> E.g., Pourable: the first sentence of the comment is ambiguous (see Alice above). The comment is further strange in that it seems to want to resolve this ambiguity, but does so in conflicting ways. On the one hand, we have "(the barrier)" in the first sentence which suggests we are talking about the source container. However the second sentence talks about the fluid.
> 
> I suggest that, since natural language use is a little ambiguous when pouring is concerned, it will be ok if the comments will be a little "strained" in terms of expression. E.g. something like the disposition of a fluid or substance which makes it possible to pour it out of a container and into or onto other objects, a role of some fluid or substance that is the patient of pouring task. Still not quite perfect as not only fluids are pourable (powder is pourable and I'm not sure it counts as a fluid).

Okay, the typo is removed and the comment is updated.

mpomarlan:
This part ", a role of some fluid or substance that is the patient of pouring task" was not intended to be in the comment for the Pourable disposition but in the comment for the role you defined (PouredObject).

Otherwise good. Change the above and I will approve&merge.

AbhijitVyas:
> This part ", a role of some fluid or substance that is the patient of pouring task" was not intended to be in the comment for the Pourable disposition but in the comment for the role you defined (PouredObject).
> 
> Otherwise good. Change the above and I will approve&merge.

Done.

mpomarlan (approved):

PR #237, opened 2022-07-01
CRAM Model and remodeling of SOMA IO

mrnolte:
This is the draft PR for the ongoing work to remodel SOMA IO and to model CRAM in SOMA.


mrnolte:
Changed to a PR, since this has gotton so big that the current changes should be reviewed.

daniel86 (commented):
I was quickly browsing through the changes. But there are too many to do it in detail. Please make individual pull requests for the different things that you did. e.g. adding labels should be one request which will be super easy to approve.

The only thing I am very concerned of that occurred to me is changing DUL.owl. I would tend not doing it due to later synchronization issues. I see you introduced Actor, and maybe also changed some existing axioms referring to Agent. Wouldn't it be enough to state external from DUL.owl that Agent is a sub-concept of Actor? could be defined in SOMA-ACT maybe?

PR #243, opened 2022-07-05
Fixed index.html and moved files

sasjonge:
I fixed the index.html to contain the correct paths. Additionally I added the subontologies to the version 1.2.0 directory and fixed the structure of the files directory.


daniel86 (changes requested):
I think the import statements must be rewritten.
e.g. SOMA-ELAN.owl imports <owl:imports rdf:resource="http://www.ease-crc.org/ont/SOMA-ACT.owl"/>
but it should be <owl:imports rdf:resource="http://www.ease-crc.org/ont/1.2.0/SOMA-ACT.owl"/>

sasjonge:
> I think the import statements must be rewritten. e.g. SOMA-ELAN.owl imports <owl:imports rdf:resource="[http://www.ease-crc.org/ont/SOMA-ACT.owl"/>](http://www.ease-crc.org/ont/SOMA-ACT.owl%22/%3E) but it should be <owl:imports rdf:resource="[http://www.ease-crc.org/ont/1.2.0/SOMA-ACT.owl"/>](http://www.ease-crc.org/ont/1.2.0/SOMA-ACT.owl%22/%3E)

I removed the ontologies from this PR so that this PR only fixes the index.html and moves the non-owl files.

mpomarlan:
Will the imported version be updated automatically? It should track the latest version, which is why there is often a convention to have a path without version numbers which goes to the most recent version, and if you need a particular version then you import from a path that contains said version.

PR #254, opened 2022-07-19
Update DUL to differentiate betweeen Agent and Actor

mrnolte:
We agreed to differentiate between types of entities with the ability to "act": Actors, and Agents, where we see the latter as more strict with the additional requirement to have some understanding of their actions.
This unfortunately requires a small change to DUL, because DUL axiomatizes Actions as having an Agent as a participant. For Communication Actions however, e.g., sending and receiving a message, we need them to "only" need an Actor to model the communication between, for example, servers.


daniel86 (changes requested):
- one thing missing is changing the comment of Action which mentions Agent explicitly. Potentially there are more instances of such comments in DUL.owl.
- I am not entirely convinced that this is the right way to go. Because the action-agent pattern includes more, in particular intentions/goals which a non-agentive agent probably does not have. this would mean we need to distinguish between Agentive and NonAgentive actions. Idea: could we introduce NonAgentiveAction in addition to DUL's Action?
- one could also see the NonAgentiveActions as processes set into motions through AgentiveActions, or?

PR #255, opened 2022-07-19
Moved Designed Tool

mrnolte:
I moved Designed Tool from SOMA-OBJ to SOMA, because we need it in SOMA-IO as Software Instance should be a subclass of that.


daniel86 (commented):
and it would not be ok to import SOMA-OBJ into SOMA.IO?

daniel86:
> From my side, I do not see a problem with that. Just thought that the current import structure follows some system to modularize the knowledge as best as possible?

it's just that the SOMA.owl file is kind of a pool of things that did not fit into the modules. So it does not really have a clearly defined scope. I was still thinking we will try to move everything in SOMA.owl into fitting modules. let's discuss this in person....

mrnolte:
Changed as discussed such that SOMA IO imports SOMA OBJ instead of moving DesignedTool

sasjonge (approved):
This seems good to me. It is just the addition of one import which is as expected.

PR #257, opened 2022-07-19
Update SOMA-SAY.owl

mrnolte:
1. I changed the IRIs and Labels of some concepts, because they were to unspecific. E.g., "Query" was a Task, but with the ongoing remodeling of SOMA IO we need to distinguish between The task of querying, a linguistic question and a formal query for some database or reasoner.  
2. I added some missing labels
3. I specified the relation to some IOlight objects, i.e., LinguisticObject and Phrase


daniel86:
looks good to me. But I wasn't involved with SOMA-SAY, maybe @mpomarlan could do the review?

mpomarlan (approved):

PR #258, opened 2022-07-19
Update SOMA-ACT.owl

mrnolte:
1. Added lots of missing labels
2. Communication Task now uses Actors instead of Agents (see #254)
a) In SOMA SAY I added Locutionary Act, which uses the stricter Agents (see #257)

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -332,7 +332,7 @@ For Manipulating, it is the movement of the hand(s) and the change in functional
                 <owl:someValuesFrom rdf:resource="http://www.ontologydesignpatterns.org/ont/dul/IOLite.owl#LinguisticObject"/>
             </owl:Restriction>
         </rdfs:subClassOf>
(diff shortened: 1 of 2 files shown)


owl/SOMA-ACT.owl
```diff
@@ -1346,6 +1376,23 @@ As such, an explicit assertion is needed to make a Situation a nonmanifested one
...
+    <!-- 
+    ///////////////////////////////////////////////////////////////////////////////////////
+    //
+    // Annotations
+    //
+    ///////////////////////////////////////////////////////////////////////////////////////
+     -->
+
+    <rdf:Description rdf:about="http://www.ease-crc.org/ont/SOMA.owl#transitionsBack">
+        <rdfs:label xml:lang="en">transitions back</rdfs:label>
```
daniel86, line 1388: this seems odd. I assume the properties are defined somewhere else? the label should be moved there

daniel86 (changes requested):
there is nothing really critical here. Some comments change 'Agent' to 'Actor' so a decision about the modification of DUL.owl should be done first.
Two comments seem to be added to the wrong file.

PR #264, opened 2022-07-22
Fix for #248

### LINKED ISSUE #248: Switch to Manchester Syntax
For reviews, it would be nice to have the ontologies and the changes in Manchester Syntax, which is designed to be human-readable.
It is also the Ontology Language used by Protégé, e.g., to define complex concepts. It might be neccessary to convert the Manchester Syntax to XML for online releases (via the CI).

mrnolte:
This is in preparation to later fix #248 when we have merged all pending .owl pull requests


sasjonge (approved):
The restructuring looks good. The resulting OWL files look correct

Issue #251, opened 2022-07-12
OntoMetrics and Webvowl not working with version 1.2.0 and current version


sasjonge:
OntoMetrics is still not working. It seems to work fine if I test it with the domain-ontologies but failes with the collapsed variant.

sasjonge:
I found out that Ontometrics, as well as Webvowl are not working with version 1.2.0 and higher. 1.1.0 is working. I will update the title of this issue accordingly.

sasjonge:
Update: I found out that removing all owl:Axioms that contain http://www.w3.org/2002/07/owl#Nothing

For example:

https://github.com/ease-crc/soma/blob/897a702b35dbe3c038ef5f213d71b1e3cd65b56f/owl/current/SOMA.owl#L5776

mrnolte:
> Update: I found out that removing all owl:Axioms that contain http://www.w3.org/2002/07/owl#Nothing
> 
> For example:
> 
> https://github.com/ease-crc/soma/blob/897a702b35dbe3c038ef5f213d71b1e3cd65b56f/owl/current/SOMA.owl#L5776

What happens if you remove those axioms? You haven't stated that 😅

sasjonge:
If i remove them, then Ontometrics and Webvowl are working correctly.

sasjonge:
I created an issue in the WebVOWL repository:

https://github.com/VisualDataWeb/WebVOWL/issues/191

mrnolte:
Is this now fixed (on our side)?

sasjonge:
Yes, it is atleast fixed in the sense that the ontologies work with both services. I have to add Ontometrics back, because some changes are necessary.

PR #265, opened 2022-07-22
Automatic GCI conversion

### LINKED ISSUE #236: Unintuitive axioms
In SOMA ACT, we have, e.g., the following axiom:
```owl
hasInitialSituation some NonmanifestedSituation SubClassOf owl:Nothing
```
`hasInitialSituation` currently has the range `Situation`.
The above axiom could be expressed by changing that range to `Situation and not NonmanifestedSituation`.
This would have the advantage that one can directly see this restriction when inspecting `hasInitialSituation` - otherwise, one has to look at the usages or at owl:Nothing.

A similar axiom can be refactored for `hasTerminalSituation`.

mrnolte:
This introduces automatic GCI conversion such that in the deployed ontology, some GCIs with owl:Nothing as super class will be rewritten to Disjointness or ObjectPropertyRange axioms. Should hopefully fix #236 and #251


mrnolte:
Please merge #248 before.

sasjonge:
I tested the resulting ontologies with Webvowl and Ontometrics and they are working. Here is the resulting Webvowl visualization:

working_soma

I will append the resulting OWL Files here, you can also see the result at my fork on the master: https://github.com/sasjonge/soma/actions/runs/2727859329

[resulting_soma.zip](https://github.com/ease-crc/soma/files/9176181/resulting_soma.zip)


@mpomarlan Can you check the content of the resulting SOMA files, if the conversed GCI's are okay?

mpomarlan:
Almost perfect!

One of the GCIs got a bad conversion. Fault of the GCI for being a bit redundantly written perhaps, but still an important GCI:

Situation and ('manifests in' some Event) and (prevents some ('manifests in' some Event)) SubClassOf owl:Nothing

This intends to say that a manifested situation cannot prevent another manifested situation. (Or in other words, either something prevented did not manifest, or what could have prevented something did not manifest.)

The redundancy is "Situation and". Because of the domain of the 'manifests in' property this is not needed and would be an easy fix for this issue. As it is, the resulting ontology contains the axiom:

disjoint classes: situation, manifests in some Event, prevents some manifests in some Event

This is a bit harsh, as it implies no situation can ever manifest.

mrnolte:
Good catch, should now be fixed

mpomarlan:
Sascha, can you make an updated version of the merged SOMA so I can check?

sasjonge:
Here is the updated version

[resulting_soma.zip](https://github.com/ease-crc/soma/files/9188960/resulting_soma.zip)

You can also find it as results of my CI here:

https://github.com/sasjonge/soma/actions/runs/2738737621

mpomarlan:
Now I see that a Situation is asserted as equivalent to something which does NOT (manifest in an event AND prevent something that manifests in an event). (I'm made sure by checking the owl text because Protege should render a few more parantheses for clarity but that's not our fault.)

I understand the subclassof direction of that equivalence. In fact, if this were a subclass axiom rather than equivalence it should have captured the original axiom exactly.

What's the justification for the opposite direction? It seems too strong, since many things will both not manifest and not prevent anything. In particular, things outside the domain of prevents and manifests will not do this, such as classes disjoint with Situations such as Concept or Description. But then it would follow these are Situations as well, which I think would then entail there are no Concepts nor Descriptions. Didn't the reasoner catch this or am I making a mistake in inference?

mpomarlan:
There's also a general class axiom that seems to have changed:

not (('has precondition' some 'Nonmanifested situation')) EquivalentTo 'manifests in' some Event

While this doesn't seem to lead to empty concepts, imo it says too much. Not everything that could manifest (ie., has all preconditions met) actually does.

mrnolte:
Sorry for the typo; I had subclassOfAxiom in mind and picked the wrong method.

mpomarlan:
Good! Maybe Sascha can generate the merged SOMA again, I'll have one more check and hopefully that's that for this.

sasjonge:
@mpomarlan Here is the updated version

[resulting_soma.zip](https://github.com/ease-crc/soma/files/9189930/resulting_soma.zip)

and the results of my CI are here:

https://github.com/sasjonge/soma/actions/runs/2739590300

mpomarlan (approved):

Issue #267, opened 2022-07-22
Generation of Vocab PDF is failing


sasjonge:
It seems that the changes to the CI somehow broke the creation of the pdf containing the vocabulary.

See

https://github.com/ease-crc/soma/runs/7471432084?check_suite_focus=true

PR #269, opened 2022-07-24
Fix for Vocabulary Pythonlatex CI Job

sasjonge:
This will fix https://github.com/ease-crc/soma/issues/267

I'm not sure why suddenly the URL became a problem, but I tested this on my master and it worked.


mrnolte (commented):
[Tested it](https://github.com/mrnolte/soma/runs/7518055504?check_suite_focus=true) and seems to fox the issue.

mrnolte (approved):

PR #270, opened 2022-07-26
Fixed Ontometrics call by moving to a POST call

sasjonge:
Ontometrics GET interface is not working anymore, therefore I switched to POST. It was necessary to move the metrics step after the "deploying the ontologies" step to have the current OWL file accessible over the URI.

Here is my successful CI run with the new metrics file:

https://github.com/sasjonge/soma/actions/runs/2740607407

Because of the move to POST, I had to add an extra step for a release, so that the metrics are generated from the released version and not the one published as "current".


sasjonge:
This PR should also finally fix https://github.com/ease-crc/soma/issues/251

daniel86 (approved):

PR #277, opened 2022-08-24
(Kaviya's PR) Kitchen, Oven, TrashContainer concepts 

mpomarlan:
@kaviyachandran  's PR, with some touchups.

I didn't change anything about her definition of Kitchen. I think the containment disposition should be pushed higher to Room, but we can discuss that later.

The containment dispositions asserted for Oven and TrashContainer turned out to not be needed, they were inheritted through these objects being DesignedContainers. To each, I added a more specific disposition.


owl/SOMA-HOME.owl
```diff
@@ -106,6 +130,30 @@
...
+                    </owl:Restriction>
+                </owl:allValuesFrom>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment>The disposition of an object to connect with some other, so that a region of space becomes fully enclosed by the object and separate from the exterior. 
+
+Often this connection is between parts of the same object.
+
+Because of the existence of a separation between some interior and exterior space regions, objects that can close are also objects that can contain.</rdfs:comment>
+    </owl:Class>
```
mrnolte, line 153: Two questions: 
1. Why do we use isQualityOf and hasQuality instead of isDispositionOf and hasDisposition?
2. ClosingDisposition -> Containment might be valid, however, does then a room have the disposition of being closeable or the door? Because, I guess, a door does not have the containment disposition, but a room does?

owl/SOMA-HOME.owl
```diff
@@ -385,6 +464,39 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#DesignedContainer"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
+                <owl:onProperty rdf:resource="http://www.ease-crc.org/ont/SOMA.owl#hasDisposition"/>
+                <owl:someValuesFrom rdf:resource="http://www.ease-crc.org/ont/SOMA-HOME.owl#ClosingDisposition"/>
+            </owl:Restriction>
+        </rdfs:subClassOf>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">An item designed to hold trash. Typically equipped with a mechanism to close or open it, to prevent smells from the trash from propagating out and insects and other pests to get in.</rdfs:comment>
+    </owl:Class>
+
```
mrnolte, line 497: No so sure about this one: A paper bin is arguably a trashContainer, but is not closeable.

mrnolte (changes requested):

mpomarlan:
Ok, I'll leave the more adventurous assertions out for now.

I'm not sure what dispositions would go for trash container beyond it being a container though, so if anyone asks that's why it's empty in its specific disposition list :)

mrnolte (approved):

PR #282, opened 2022-09-26
Perception Capability with dispositions

AbhijitVyas:


mrnolte:
There has been [work by me to also introduce capabilities](https://github.com/mrnolte/soma/tree/master), together with a new property, "affords executor", besides "affords bearer" and "affords trigger". I would like to discuss the proposed changes on Thursday.

PR #283, opened 2022-09-27
Removed SOMA from SOMA-HOME

sasjonge:
At the moment we would have duplicate triples in KnowRob if they appear in SOMA and SOMA-HOME. It also seems to make more sense to me to have SOMA-HOME import SOMA instead of the current merging. 

I wanted to create a SOMA-HOME-bundled file but just adding it to the application.yaml would throw an error.


sasjonge:
An example run of this change can be found here:

https://github.com/sasjonge/soma/blob/gh-pages/owl/current/SOMA-HOME.owl

daniel86 (approved):

Issue #288, opened 2022-10-28
Change SOMA-OBJ IRIs to SOMA


mrnolte:
Somce entities use namespaces other than dul or soma, e.g., soma-obj. We have to clean that up.

daniel86:
I did this in the Prolog script to avoid that the bundled version has different IRIs. Protege does not work well together with different OWL files sharing the same namespace, in my experience.

PR #289, opened 2022-11-16
Added concepts that were needed to model new IAI apartment.

daniel86:
I only added a few concepts, including comments.
Some axioms will follow in another pull request.

Not sure though why there is so much refactoring ongoing. I edited with Protege 5.5.0

### DIFF
--- a/owl/SOMA-HOME.owl
+++ b/owl/SOMA-HOME.owl
@@ -12,59 +12,43 @@ Ontology(<http://www.ease-crc.org/ont/SOMA-HOME.owl>
 Import(<http://www.ease-crc.org/ont/SOMA.owl>)
 Annotation(rdfs:comment "SOMA-HOME defines the concepts that are specific to home, e.g. kitchen environment, household items.")
 
-Declaration(Class(:Box))
-Declaration(Class(:BreadKnife))
-Declaration(Class(:BreakfastPlate))
-Declaration(Class(:Carafe))
-Declaration(Class(:CerealBox))
-Declaration(Class(:ClosingDisposition))
-Declaration(Class(:CoffeeCarafe))
-Declaration(Class(:Cup))
-Declaration(Class(:DessertFork))
-Declaration(Class(:DinnerPlate))
-Declaration(Class(:Dispenser))
-Declaration(Class(:Glass))
-Declaration(Class(:JamJar))
-Declaration(Class(:Jar))
-Declaration(Class(:Kitchen))
-Declaration(Class(:KitchenKnife))
-Declaration(Class(:MilkBottle))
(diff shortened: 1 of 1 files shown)


owl/SOMA-HOME.owl
```diff
@@ -12,59 +12,43 @@ Ontology(<http://www.ease-crc.org/ont/SOMA-HOME.owl>
...
-Declaration(Class(:TableSpoon))
-Declaration(Class(:TeaSpoon))
-Declaration(Class(:TemperingByCooling))
-Declaration(Class(:TemperingByHeating))
-Declaration(Class(:TrashContainer))
-Declaration(Class(:WaterBottle))
-Declaration(Class(:WaterGlass))
-Declaration(Class(:WineBottle))
-Declaration(Class(:WineGlass))
+Declaration(Class(<http://www.ease-crc.org/ont/SOMA-HOME.owl#Armchair>))
```
mpomarlan, line 15: I'm a bit confused by the prefixes situation.

New concepts appear here with SOMA-HOME.owl in the prefix. At the same time, something like NutellaJar (seems too specific to be in SOMA proper) uses SOMA.owl as prefix. This is not an issue with this PR as such, rather it is something to keep in mind, perhaps by opening a github issue.

mpomarlan (approved):

mpomarlan:
What's up with the SOMA Vocabulary step?

daniel86:
> What's up with the SOMA Vocabulary step?

to me it looks like the CI step cannot handle the new format. But I haven't looked deeper into it

owl/SOMA-HOME.owl
daniel86: There is a related issue here: https://github.com/ease-crc/soma/issues/288

daniel86:
>

daniel86:
> > What's up with the SOMA Vocabulary step?
> 
> to me it looks like the CI step cannot handle the new format. But I haven't looked deeper into it.
> 
> seems owl2ready is used: https://github.com/ease-crc/soma/blob/master/docs/vocab/owl_reader.py#L5

"Owlready2 currently reads the following file format: RDF/XML, OWL/XML, NTriples. The file format is automatically detected." they write here https://owlready2.readthedocs.io/en/latest/onto.html

daniel86:
> What's up with the SOMA Vocabulary step?

yeah it looks like owl2ready cannot handle functional syntax, which is the format of the file passed to it here: https://github.com/ease-crc/soma/blob/master/docs/vocab/SOMA-vocabulary.tex#L21
However, it looks like the wrong file is passed, i.e., not the bundled version. So I guess a fix would be to instead supply the bundled version in XML format to owl2ready.

PR #299, opened 2023-04-21
Added affects-relationship

mrnolte:
Introduced a more general version of `causes`: `affects` (and its inverse `is affected by`) to express that a variation in the course or outcome of the the subject (the affector) would have resulted in a variation in the object (the affectee), e.g., a planning task that sets parameters such as goal position affects the subsequently executed pick-and-place task that uses that parameter.

This was not captured by `causes`: In the above scenario, the pick-and-place action was not necessarily caused by the planning task, e.g., when otherwise a default value would have been used by the agent.

### DIFF
--- a/owl/SOMA-ACT.owl
+++ b/owl/SOMA-ACT.owl
@@ -200,7 +200,7 @@ Declaration(NamedIndividual(SOMA:ExecutionState_Succeeded))
 
 # Object Property: SOMA:affects (affects)
 
-AnnotationAssertion(rdfs:comment SOMA:affects "Simple relationship between two actions to express that a variation in the course or outcome of the the object (the affectee) would have resulted in a variation in the subject (the affector), e.g., a planning task that sets parameters such as goal position affects the subsequently executed pick-and-place task that uses that parameter.")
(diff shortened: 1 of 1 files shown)


owl/SOMA-ACT.owl
```diff
@@ -196,6 +198,13 @@ Declaration(NamedIndividual(SOMA:ExecutionState_Succeeded))
 #   Object Properties
 ############################
 
+# Object Property: SOMA:affects (affects)
+
+AnnotationAssertion(rdfs:comment SOMA:affects "Simple relationship between two actions to express that a variation in the course or outcome of the the object (the affectee) would have resulted in a variation in the subject (the affector), e.g., a planning task that sets parameters such as goal position affects the subsequently executed pick-and-place task that uses that parameter.")
```
mpomarlan, line 203: Apologies for the nitpick but it seems the comment should say, a change in the subject (affector) results in a change of the object (affectee). Otherwise the PR is good to go!

mpomarlan (changes requested):

owl/SOMA-ACT.owl
mrnolte: Fixed this and the comments of its inverse!

PR #308, opened 2023-10-23
Extend preference concepts

ayden175:
Builds upons PR https://github.com/ease-crc/soma/pull/297 (Soma-Agent) to add the rest of the preference concepts. Some of the changes from Soma-Agent seem to not directly fit with some concepts of the preference model from the paper:

####  Preference
- currently has the relation `'is quality of' only Agent`
- however, paper proposed `'is preference of' only Agent`
-> should `'is preference of'` be added and used instead? (as a subrelation of `'is quality of'`?)

#### Preference Order
- Soma-Agent added `Predilection`, but that seems to be slightly different (orders situations instead of discriptions)
-> should `Predilection` be kept as well, or replaced by `Preference Order`?

#### Ordered Element
- Soma-Agent added `Order Item`, which seems to do the same things as far as I can tell
- `Order Item` uses `hasMember` instead of the in the paper proposed `encapsulates`
-> should `Order Item` be kept, or (partially) replaced by `Ordered Element` or `encapsulates`?

### DIFF
--- a/owl/SOMA-AGENT.owl
+++ b/owl/SOMA-AGENT.owl
@@ -18,19 +18,39 @@ Declaration(Class(SOMA:AgentRole))
 Declaration(Class(SOMA:Choice))
 Declaration(Class(SOMA:Option))
 Declaration(Class(SOMA:Order))
-Declaration(Class(SOMA:OrderItem))
+Declaration(Class(SOMA:OrderedElement))
 Declaration(Class(SOMA:Predilection))
 Declaration(Class(SOMA:Preference))
 Declaration(Class(SOMA:PreferenceOrder))
 Declaration(Class(SOMA:PreferenceRegion))
 Declaration(Class(SOMA:Selecting))
 Declaration(Class(SOMA:Singleton))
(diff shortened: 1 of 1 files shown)


mpomarlan (approved):

mpomarlan:
I would say to obey the paper as much as possible. With this rule of thumb, answers:

1) Yes, use "is preference of" which is a subproperty of "is quality of"
2) At the moment I do not care much one way or the other. We can keep Predilection for now.
3) Who else is using OrderedItem? If only the preference model, then let us stick to the paper names and only keep OrderedElement and its encapsulates property.

mpomarlan:
Hello there, what else needs doing here? We should close this and release soon.

ayden175:
I somehow missed the first comment, sorry! I've adjusted the changes accordingly. OrderItem was not used by anything else, so I renamed it and let it use encapsulates instead of hasMember. If the new changes look ok, the PR would be ready to be merged now.

mpomarlan (approved):

PR #314, opened 2024-05-15
Documentation Fix

### LINKED ISSUE #305: Documentation is broken
The SOMA vocabulary job fails. Seems like we need to convert the functional syntax to a format that owlready2 can read for that.

mrnolte:
This PR fixes #305. Multiple things had to be touched, as the switch from RDF/XML to functional syntax was not the only issue.

1.  `documentation.yaml`, `evaluation.yaml` and `SOMA-vocabulary.tex` have been adapted so that the Evaluation workflow does not use the raw SOMA files, but the ones generated by the evaluation (these are uploaded as artifacts).
2. Two annotations in SOMA-AGENT were incorrectly labels instead of comments; the documentation did not like the contained line breaks.
3. `owl-reader.py` now correctly escapes latex special characters such as `_`.
4. SOMA-ACT had one class, `NonmanifestedSituation`, which used the incorrect separator `/` instead of `#` between namespace and fragment.

The Release Notes have been updated, including a sentence about USD.owl.


.github/workflows/documentation.yml
```diff
@@ -1,10 +1,9 @@
 name: Documentation
 on:
-  push:
-    paths:
-      - 'owl/**'
-      - 'docs/**'
-      - '.github/**'
+  workflow_run:
```
daniel86, line 3: I am not sure about this change. So basically before the trigger was *push* operation that changes one of the listed paths. If I understand correctly you changed it to "whenever the 'evaluation' workflow was triggered successfully" which is not right because the workflow is executed for pull requests which are not yet accepted! I think with this block the docu would update before the pull request is accepted

daniel86 (changes requested):

.github/workflows/documentation.yml
daniel86: I think it is ok to revert the changes back such that the docu is only generated on push again.

mrnolte: Unfortunately, this is not a solution: The Documentation *must* run *after* the Evaluation, because the Evaluation downloads an artifact generated by the Evaluation. With your approach, however, both would run in parallel - there would be no triggering run, and we could not reference where to get the artifact from (see documentation.yaml line 25).
What we can do, however, is to ignore both jobs if the triggering run was itself not triggered by a push. I added this now via `if: ${{ github.event.workflow_run.event == 'push' }}`, see documentation.yaml line 10 and 47.

daniel86 (approved):

PR #316, opened 2024-07-12
Added classes for motion processes

sasjonge:
In this PR I add the concepts of motion process and motion plan, with corresponding properties.


mrnolte (changes requested):
I think the introduced changes look good, but please also add a small note to the release notes about the additions.

mrnolte (approved):

PR #320, opened 2025-05-12
Initial version of PyCRAM Designators in SOMA

mrnolte:
Here is a minimal addition of PyCRAM Designators to SOMA via SOMA-PYCRAM. Does it make sense? And, especially, does it make sense to put it into the main repository?


owl/catalog-v001.xml
```diff
@@ -3,14 +3,15 @@
     <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/DUL.owl" uri="DUL.owl"/>
     <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA.owl" uri="SOMA.owl"/>
     <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA-ACT.owl" uri="SOMA-ACT.owl"/>
-	<uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA-AGENT.owl" uri="SOMA-AGENT.owl"/>
-    <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA-ALL.owl" uri="SOMA-ALL.owl"/>
+    <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA-AGENT.owl" uri="SOMA-AGENT.owl"/>
+    <uri id="User Edited Redirect" name="http://www.ease-crc.org/ont/SOMA-All.owl" uri="SOMA-ALL.owl"/>
```
mpomarlan: SOMA-ALL became SOMA-All in the name. Is this safe or will it break anything?

owl/SOMA-PYCRAM.owl
```diff
@@ -0,0 +1,66 @@
+Prefix(:=<http://www.semanticweb.org/robin/ontologies/2025/4/untitled-ontology-8/>)
```
mpomarlan: A minor nitpick but the : prefix should be something SOMA related, rather than a Protege generated-default.

owl/SOMA-PYCRAM.owl
```diff
@@ -0,0 +1,66 @@
...
+
+
+
+############################
+#   Classes
+############################
+
+# Class: <http://www.ease-crc.org/ont/SOMA.owl#PyCramActionDesignator> (PyCRAM Action Designator)
+
+AnnotationAssertion(rdfs:comment <http://www.ease-crc.org/ont/SOMA.owl#PyCramActionDesignator> "A PyCRAM Designator expressing a Plan for a Task.")
```
mpomarlan: AFAIK action designators are not exactly plans. They specify what, not how. 

If possible, I would go with any "official" definitions for pycram action designators here. E.g., what does the pycram documentation describe them as? If it turns out the pycram doc describes them as plans, it is itself likely wrong however and we would need a different text for them.

owl/SOMA-PYCRAM.owl
```diff
@@ -0,0 +1,66 @@
...
+AnnotationAssertion(rdfs:label <http://www.ease-crc.org/ont/SOMA.owl#PyCramActionDesignator> "PyCRAM Action Designator")
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#PyCramActionDesignator> <http://www.ease-crc.org/ont/SOMA.owl#PyCramDesignator>)
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#PyCramActionDesignator> ObjectSomeValuesFrom(<http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#expresses> <http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Plan>))
```
mpomarlan: See above. Worth discussing, because action designators, at least in Lisp CRAM, rather expressed goals than plans.

owl/SOMA-PYCRAM.owl
```diff
@@ -0,0 +1,66 @@
...
+
+# Class: <http://www.ease-crc.org/ont/SOMA.owl#PyCramDesignator> (PyCRAM Designator)
+
+AnnotationAssertion(rdfs:comment <http://www.ease-crc.org/ont/SOMA.owl#PyCramDesignator> "A PyCRAM designator is a structured data object (possibly existing as a file or in memory only) that concretely encodes an (underspecified) description (dul:Description) used by the CRAM framework to resolve objects, actions, motions, etc. at runtime.
+
+As structured data, it conveys:
+- For object lookups: a dul:Configuration of feature‐roles (e.g., type, color, shape).
+- For actions and motions: a dul:Method that prescribes how to carry out an underspecified task or movement toward a goal.
```
mpomarlan: true for motion. As discussed above, at least this was not the intention for the action designators. The reason was that several different actions may be appropriate for achieving a goal, and thus the idea was to specify a common core -- the goal to be pursued -- and then let the designator resolution process to fill in the details of how the action is to be performed. Designators can add more constraints to this description, all the way down to fully specifying what actions to use and with what parameters, but this is not intended as a typical usecase.

owl/SOMA-PYCRAM.owl
```diff
@@ -0,0 +1,66 @@
...
+AnnotationAssertion(rdfs:label <http://www.ease-crc.org/ont/SOMA.owl#PyCramObjectDesignator> "PyCRAM Object Designator")
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#PyCramObjectDesignator> <http://www.ease-crc.org/ont/SOMA.owl#PyCramDesignator>)
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#PyCramObjectDesignator> ObjectSomeValuesFrom(<http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#expresses> <http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#Configuration>))
+
+# Class: <http://www.ease-crc.org/ont/SOMA.owl#Resolving_of_PyCRAM_Designators> (Resolving a PyCRAM Designator)
+
+AnnotationAssertion(rdfs:label <http://www.ease-crc.org/ont/SOMA.owl#Resolving_of_PyCRAM_Designators> "Resolving a PyCRAM Designator")
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#Resolving_of_PyCRAM_Designators> <http://www.ease-crc.org/ont/SOMA.owl#DerivingInformation>)
+SubClassOf(<http://www.ease-crc.org/ont/SOMA.owl#Resolving_of_PyCRAM_Designators> ObjectSomeValuesFrom(<http://www.ease-crc.org/ont/SOMA.owl#isTaskOfInputRole> <http://www.ease-crc.org/ont/SOMA.owl#PyCramDesignator>))
```
mpomarlan: What are the domains and ranges of the isTaskOfInput/OutputRole?

owl/catalog-v001.xml
mrnolte: This was because the prefix in the file SOMA-ALL was actually SOMA-All, I fixed this

owl/SOMA-PYCRAM.owl
mrnolte: Fixed this!

owl/SOMA-PYCRAM.owl
mrnolte: Well, I'd argue that DUL does not offer anything better than a `DUL:Plan` for this, unfortunately. I quote the documentation (https://pycram.readthedocs.io/en/latest/designators.html):

> Action designators can be seen as higher level plans that include failure handling and parametrization.

Ontologically, they are of course not the plan itself, merely a structured representation of said plan; i.e., they must be a `DUL:InformationObject` (more specifically, they are `SOMA:StructuredText`, but this does not matter for this argument). As a `DUL:InformationObject`, they `DUL:expresses` some `DUL:SocialObject`. This is true for all PyCRAM Designators (!). Now, for Action Designators we may consider different `SocialObject`s, e.g., `DUL:Task`, `DUL:Plan`, and `DUL:Method`. For `DUL:Task`, we would need a `DUL:Description` that `DUL:defines` it. This to me is unintuitive, as while a DUL:Task may have subtasks, it is not structured - the structure stems from a plan (as far as I understand it). For a` DUL:Plan`, DUL says 

> A Plan is a Description of some actions to be executed by agents in a certain way, with certain parameters

This to me sound exactly to what Designators are. `DUL:Method` seems wrong to me, because it more vague, i.e., DUL says

> It is different from a Plan, because plans could be carried out in order to follow a method, but a method can be followed by executing alternative plans.

So, when your saying that

> If it turns out the pycram doc describes them as plans, it is itself likely wrong however and we would need a different text for them.

what do you actually mean by that? What do you think do they express?

owl/SOMA-PYCRAM.owl
mrnolte: Ah I now see; however, we have that `DUL:Plan DUL:hasComponent some DUL:Goal`, i.e., we have something more general than the goal itself.

owl/SOMA-PYCRAM.owl
mrnolte: Now, following your line of argumentation, I believe we must look at two different things here, actually. So, an unresolved action designator would express a `DUL:Method`, while a resolved action designator would express a `DUL:Plan`?

owl/SOMA-PYCRAM.owl
mrnolte: Ah damn, I forgot the Role inbetween. Will change that, good catch!

mrnolte:
I now added more specific classes to the model, which divide the underspecified description (PyCRAM Designator Description) and the specified description (PyCRAM Designator Referent). @mpomarlan does this now resolve your issues with the modeling?

As a side note, there is yet another issue, in which the description itself, e.g., "a blue cup" must be somehow represented. This is problematic in that OWL does not allow nesting of expressions. I fear that we must reify the concepts and relations.

mpomarlan:
Could you expand on the blue cup comment?

For the rest, it turns out we have 2, soon to be 3, PRs incoming about designators. Yours, Sascha's, and Alina's. It would be good I think to harmonize these in a SOMA meeting, e.g. this Friday.

PR #321, opened 2025-05-14
CROMA - Draft

hawkina:
This is a draft version of CROMA.owl, which is supposed to represent CRAM-Designators and connect them to SOMA. Please feel free to move the newly added classes to other subclasses (I wasn't sure about where they should go but just wanted all the classes to exist already) or notify me if anything is unclear. 
Let me know if I forgot to add annotations to things.

### DIFF
--- a/owl/CROMA.owl
+++ b/owl/CROMA.owl
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <rdf:RDF xmlns="http://www.ease-crc.org/ont/CROMA.owl#"
-     xml:base="http://www.ease-crc.org/ont/CROMA.owl"
+     xml:base="https://raw.githubusercontent.com/hawkina/soma/refs/heads/soma-cram/owl/CROMA.owl"
      xmlns:dc="http://purl.org/dc/elements/1.1/"
      xmlns:owl="http://www.w3.org/2002/07/owl#"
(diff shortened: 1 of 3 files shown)


owl/CROMA.owl
```diff
@@ -0,0 +1,438 @@
...
+    </owl:Class>
+    
+
+
+    <!-- http://www.ease-crc.org/ont/CROMA.owl#CRAM_Designator -->
+
+    <owl:Class rdf:about="http://www.ease-crc.org/ont/CROMA.owl#CRAM_Designator">
+        <rdfs:subClassOf rdf:resource="http://www.ease-crc.org/ont/CROMA.owl#Designator"/>
+        <rdfs:subClassOf>
+            <owl:Restriction>
```
mpomarlan, line 170: This restriction should say that a CRAM designator is resolved_to only a CRAM designator. Not all designators are resolved, but those that are can only be resolved to designators.

mpomarlan:
Would be nice to add some axioms about the resolved_to/_from relations, specifically stuff like action designator resolved being resolved_from only action designator description (or whatever the least resolved one is), and action designator performed being resolved_from only action designator resolved. If there are some other types of designator description / resolved pairs, similar axioms should be there as well.

